PR #431, opened 2015-01-05
QAART-480

ludwikkazmierczak:


ludwikkazmierczak:
Passed three times in a row - http://qa-s2.wikia-prod:8080/view/parametersTesting/job/test-group/795/

PR #432, opened 2015-01-05
QAART-457 Address Sonar's Major issue: Local variable and method paramet...

rochan:
...er names should comply with a naming convention.

The following code snippet illustrates this rule when the regular expression value is "^[a-z][a-zA-Z0-9]*$"


ludwikkazmierczak:
LGTM

PR #434, opened 2015-01-06
QAART-443 Addressing Sonar's major issue: Switch statements should end w...

rochan:
...ith a default case


ludwikkazmierczak:
LGTM

PR #435, opened 2015-01-07
QAART-470: Removed the dependency of org.json

rochan:
As Selenium webdriver no longer supports this package as or 2.4x


ludwikkazmierczak:
LGTM

PR #452, opened 2015-01-14
Deleted layout choosing

ludwikkazmierczak:


rochan:
@ludwikkazmierczak I think the problem with removing createBlankPage is that there will be no way to exit out of the CK/Src editor without publishing an article. With blank layout, the test could just click publish and it would bring you back to the article page versus dealing with the alert modal if using standard layout.

PR #456, opened 2015-01-14
UC-190-1 adding testcases for checking for FB button on the page

bekcunning:
This test checks that FB button exists on the following pages:
-Login drop down
-Special:UserSignup
-Special:UserLogin
-Special:Preferences for logged-in user
-Forced Login modal dialog

### DIFF
--- a/src/test/java/com/wikia/webdriver/pageobjectsfactory/componentobject/dropdowncomponentobject/DropDownComponentObject.java
+++ b/src/test/java/com/wikia/webdriver/pageobjectsfactory/componentobject/dropdowncomponentobject/DropDownComponentObject.java
@@ -195,7 +195,4 @@ public class DropDownComponentObject extends WikiBasePageObject {
 		);
 	}
 
-	public void verifyFBButtonVisible() {
-		Assertion.assertTrue(checkIfElementOnPage(formConnectWithFbButton));
-	}
 }
--- a/src/test/java/com/wikia/webdriver/pageobjectsfactory/componentobject/dropdowncomponentobject/DropDownComponentObject.java
+++ b/src/test/java/com/wikia/webdriver/pageobjectsfactory/componentobject/dropdowncomponentobject/DropDownComponentObject.java
@@ -194,4 +194,8 @@ public class DropDownComponentObject extends WikiBasePageObject {
 			true
 		);
 	}
+
+	public void verifyFBButtonVisible() {
+		Assertion.assertTrue(checkIfElementOnPage(formConnectWithFbButton));
+	}
 }
--- a/src/test/java/com/wikia/webdriver/testcases/facebooktests/FacebookButtonTests.java
+++ b/src/test/java/com/wikia/webdriver/testcases/facebooktests/FacebookButtonTests.java
@@ -14,7 +14,6 @@ import org.testng.annotations.Test;
 /* 
  * Check for facebook button on the page
  */
-
 public class FacebookButtonTests extends NewTestTemplate {
 
 	Credentials credentials = config.getCredentials();
@@ -62,3 +61,4 @@ public class FacebookButtonTests extends NewTestTemplate {
 	}
 
 }
+
--- a/src/test/java/com/wikia/webdriver/testcases/facebooktests/FacebookButtonTests.java
+++ b/src/test/java/com/wikia/webdriver/testcases/facebooktests/FacebookButtonTests.java
@@ -1,7 +1,5 @@
 package com.wikia.webdriver.testcases.facebooktests;
 
-import org.testng.annotations.Test;
-
 import com.wikia.webdriver.common.properties.Credentials;
 import com.wikia.webdriver.common.templates.NewTestTemplate;
 import com.wikia.webdriver.pageobjectsfactory.componentobject.dropdowncomponentobject.DropDownComponentObject;
@@ -11,6 +9,7 @@ import com.wikia.webdriver.pageobjectsfactory.pageobject.special.SpecialNewFiles
 import com.wikia.webdriver.pageobjectsfactory.pageobject.special.login.SpecialUserLoginPageObject;
 import com.wikia.webdriver.pageobjectsfactory.pageobject.special.preferences.PreferencesPageObject;
 import com.wikia.webdriver.pageobjectsfactory.pageobject.special.preferences.PreferencesPageObject.tabNames;
+import org.testng.annotations.Test;
 
 /* 
  * Check for facebook button on the page
--- a/src/test/java/com/wikia/webdriver/testcases/facebooktests/FacebookButtonTests.java
+++ b/src/test/java/com/wikia/webdriver/testcases/facebooktests/FacebookButtonTests.java
@@ -62,4 +62,3 @@ public class FacebookButtonTests extends NewTestTemplate {
 
 }
 
-
--- a/src/test/java/com/wikia/webdriver/testcases/facebooktests/FacebookButtonTests.java
+++ b/src/test/java/com/wikia/webdriver/testcases/facebooktests/FacebookButtonTests.java
@@ -1,0 +1,79 @@
+package com.wikia.webdriver.testcases.facebooktests;
+
+import org.testng.annotations.Test;
+
+import com.wikia.webdriver.common.properties.Credentials;
+import com.wikia.webdriver.common.templates.NewTestTemplate;
+import com.wikia.webdriver.pageobjectsfactory.componentobject.dropdowncomponentobject.DropDownComponentObject;
+import com.wikia.webdriver.pageobjectsfactory.pageobject.WikiBasePageObject;
+import com.wikia.webdriver.pageobjectsfactory.pageobject.signup.SignUpPageObject;
+import com.wikia.webdriver.pageobjectsfactory.pageobject.special.SpecialNewFilesPageObject;
+import com.wikia.webdriver.pageobjectsfactory.pageobject.special.login.SpecialUserLoginPageObject;
+import com.wikia.webdriver.pageobjectsfactory.pageobject.special.preferences.PreferencesPageObject;
+import com.wikia.webdriver.pageobjectsfactory.pageobject.special.preferences.PreferencesPageObject.tabNames;
+
+/* 
+ * Check for facebook button on the page
+ */
+
+public class FacebookButtonTests extends NewTestTemplate {
+
+	Credentials credentials = config.getCredentials();
+
+	// check for FB button on login drop down
+
+	@Test(groups = { "FBButton_001", "FacebookButton" })
+	public void FBButton_001_DropDownButton_Visible() {
+		DropDownComponentObject dropDown = new DropDownComponentObject(driver);
+		dropDown.openDropDown();
+		dropDown.verifyFBButtonVisible();
+	}
+
+	// check for FB button on sign up page
+
+	@Test(groups = { "FBButtont_002", "FacebookButton" })
+	public void FBButton_002_SignUpButton_Visible() {
+		WikiBasePageObject base = new WikiBasePageObject(driver);
+		SignUpPageObject signUpPage = base.openSpecialSignUpPage(wikiURL);
+		signUpPage.verifyFBButtonVisible();
+	}
+
+	// check for FB button on login page
+
+	@Test(groups = { "FBButton_003", "FacebookButton" })
+	public void FBButton_003_LoginButton_Visible() {
+		WikiBasePageObject base = new WikiBasePageObject(driver);
+		SpecialUserLoginPageObject login = base.openSpecialUserLogin(wikiURL);
+		login.verifyFBButtonVisible();
+
+	}
+
+	// check for FB button on forced login dialog
+
+	@Test(groups = { "FBButton_004", "FacebookButton" })
+	public void FBButton_004_ForcedLoginButton_Visible() {
+		WikiBasePageObject base = new WikiBasePageObject(driver);
+		SpecialNewFilesPageObject specialPage = base
+				.openSpecialNewFiles(wikiURL);
+		specialPage.verifySpecialPage();
+		specialPage.addPhoto();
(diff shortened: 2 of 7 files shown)


src/test/java/com/wikia/webdriver/testcases/facebooktests/FacebookButtonTests.java
```diff
@@ -0,0 +1,79 @@
...
+		dropDown.openDropDown();
+		dropDown.logInCookie(credentials.userName, credentials.password,
+				wikiURL);
+		PreferencesPageObject prefsPage = base
+				.openSpecialPreferencesPage(wikiURL);
+		prefsPage.selectTab(tabNames.FACEBOOK);
+		prefsPage.verifyFBButtonVisible();
+	}
+
+}
```
rochan: Need a newline for EOF

src/test/java/com/wikia/webdriver/testcases/facebooktests/FacebookButtonTests.java
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	// check for FB button on Preferences->Facebook Connect tab
+
+	@Test(groups = { "FBButton_005", "FacebookButton" })
+	public void FBButton_005_PrefsButton_Visible() {
+		WikiBasePageObject base = new WikiBasePageObject(driver);
+		DropDownComponentObject dropDown = new DropDownComponentObject(driver);
+		dropDown.openDropDown();
+		dropDown.logInCookie(credentials.userName, credentials.password,
```
rochan: you could have this all in one line

src/test/java/com/wikia/webdriver/testcases/facebooktests/FacebookButtonTests.java
```diff
@@ -0,0 +1,79 @@
...
+				wikiURL);
+		PreferencesPageObject prefsPage = base
```
rochan: you could have this all in one line

src/test/java/com/wikia/webdriver/testcases/facebooktests/FacebookButtonTests.java
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	// check for FB button on login page
+
+	@Test(groups = { "FBButton_003", "FacebookButton" })
+	public void FBButton_003_LoginButton_Visible() {
+		WikiBasePageObject base = new WikiBasePageObject(driver);
+		SpecialUserLoginPageObject login = base.openSpecialUserLogin(wikiURL);
+		login.verifyFBButtonVisible();
+
```
rochan: remove blank line

src/test/java/com/wikia/webdriver/testcases/facebooktests/FacebookButtonTests.java
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	// check for FB button on forced login dialog
+
+	@Test(groups = { "FBButton_004", "FacebookButton" })
+	public void FBButton_004_ForcedLoginButton_Visible() {
+		WikiBasePageObject base = new WikiBasePageObject(driver);
+		SpecialNewFilesPageObject specialPage = base
```
rochan: you could have this all in one line

src/test/java/com/wikia/webdriver/testcases/facebooktests/FacebookButtonTests.java
```diff
@@ -0,0 +1,79 @@
...
+
+/* 
+ * Check for facebook button on the page
+ */
+
+public class FacebookButtonTests extends NewTestTemplate {
+
+	Credentials credentials = config.getCredentials();
+
+	// check for FB button on login drop down
```
drets: Are comments required here?
From my point of view the method name tells us about purposes of the test.

src/test/java/com/wikia/webdriver/testcases/facebooktests/FacebookButtonTests.java
```diff
@@ -0,0 +1,79 @@
...
+		WikiBasePageObject base = new WikiBasePageObject(driver);
+		DropDownComponentObject dropDown = new DropDownComponentObject(driver);
+		dropDown.openDropDown();
+		dropDown.logInCookie(credentials.userName, credentials.password,
+				wikiURL);
+		PreferencesPageObject prefsPage = base
+				.openSpecialPreferencesPage(wikiURL);
+		prefsPage.selectTab(tabNames.FACEBOOK);
+		prefsPage.verifyFBButtonVisible();
+	}
```
drets: In general, LGTM!
I just suggest to use more the [Fluent interface approach] (http://en.wikipedia.org/wiki/Fluent_interface) for increasing readability of the tests.

src/test/java/com/wikia/webdriver/testcases/facebooktests/FacebookButtonTests.java
```diff
@@ -0,0 +1,79 @@
+package com.wikia.webdriver.testcases.facebooktests;
+
+import org.testng.annotations.Test;
```
jsutterfield: The convention I've seen elsewhere is to include this line directly after all other imports, like [this](https://github.com/Wikia/selenium-tests/blob/master/src/test/java/com/wikia/webdriver/testcases/venustests/TestTablesOnVenus.java#L7). We should probably do the same here to be consistent.

src/test/java/com/wikia/webdriver/pageobjectsfactory/componentobject/dropdowncomponentobject/DropDownComponentObject.java
```diff
@@ -194,4 +194,8 @@ public void verifyMessageAboutNewPassword(String userName) {
 			true
 		);
 	}
+
+	public void verifyFBButtonVisible() {
```
jsutterfield: This ```verifyFBButtonVisible()``` here, as well as the other ```verifyFBButtonVisible()``` methods found inside SignUpPageObject.java, SpecialUserLoginPageObject.java, SpecialUserLoginPageObject.java, and PreferencesPageObject.java actually aren't necessary.

The reason is that all of these classes inherit from the WikiBasePageObject.java class. Since we have that method defined there, all other classes which inherit from it have that method as well. So you can go ahead and remove all these extra ```verifyFBButtonVisible()``` definitions except for the one found in WikiBasePageObject.java.

src/test/java/com/wikia/webdriver/pageobjectsfactory/componentobject/dropdowncomponentobject/DropDownComponentObject.java
```diff
@@ -194,4 +194,8 @@ public void verifyMessageAboutNewPassword(String userName) {
...
+		Assertion.assertTrue(checkIfElementOnPage(formConnectWithFbButton));
```
jsutterfield: When rely on just the base class version of ```verifyFBButtonVisible()``` though, we'll have to make sure that all subclasses use the same name for the instance variable it checks. Here we should rename ```formConnectWithFbButton``` to ```facebookSignUpButton``` on this line, as well as [here](https://github.com/Wikia/selenium-tests/blob/master/src/test/java/com/wikia/webdriver/pageobjectsfactory/componentobject/dropdowncomponentobject/DropDownComponentObject.java#L42) and [here](https://github.com/Wikia/selenium-tests/blob/master/src/test/java/com/wikia/webdriver/pageobjectsfactory/componentobject/dropdowncomponentobject/DropDownComponentObject.java#L131).

src/test/java/com/wikia/webdriver/testcases/facebooktests/FacebookButtonTests.java
jsutterfield: I could go either way in terms of these comments. I kinda like them, but the methods are well named so they may be redundant. If we do decide to keep them though, we should remove the space between the comment and method. eg 
```java
    // check for FB button on login drop down
 	@Test(groups = { "FBButton_001", "FacebookButton" })
	public void FBButton_001_DropDownButton_Visible() {
		DropDownComponentObject dropDown = new DropDownComponentObject(driver);
		dropDown.openDropDown();
		dropDown.verifyFBButtonVisible();
```
up to you @bekcunning.

src/test/java/com/wikia/webdriver/testcases/facebooktests/FacebookButtonTests.java
```diff
@@ -0,0 +1,79 @@
...
+import com.wikia.webdriver.pageobjectsfactory.pageobject.signup.SignUpPageObject;
+import com.wikia.webdriver.pageobjectsfactory.pageobject.special.SpecialNewFilesPageObject;
+import com.wikia.webdriver.pageobjectsfactory.pageobject.special.login.SpecialUserLoginPageObject;
+import com.wikia.webdriver.pageobjectsfactory.pageobject.special.preferences.PreferencesPageObject;
+import com.wikia.webdriver.pageobjectsfactory.pageobject.special.preferences.PreferencesPageObject.tabNames;
+
+/* 
+ * Check for facebook button on the page
+ */
+
```
jsutterfield: You can remove this space.

jsutterfield:
Looks good overall @bekcunning. Very well organized and clean. Just a few comments. Good work :)

src/test/java/com/wikia/webdriver/testcases/facebooktests/FacebookButtonTests.java
bekcunning: Can you elaborate?

src/test/java/com/wikia/webdriver/testcases/facebooktests/FacebookButtonTests.java
bekcunning: Ok, I removed the comments.  It's a working process thing that helps me outline my work before completing but I see how it is redundant.

src/test/java/com/wikia/webdriver/pageobjectsfactory/componentobject/dropdowncomponentobject/DropDownComponentObject.java
bekcunning: @jsutterfield I had the same instinct but the problem is that the name of the button is different depending on where I am looking for it -- so the one on the drop down has a different identifier than the one on the login page, which is different than the one on the login modal.  So I suppose I could define the function in WikiBasePageObject and then assign the variable in the specific page class?  I will try this.

src/test/java/com/wikia/webdriver/testcases/facebooktests/FacebookButtonTests.java
rochan: You could put this all in one line
```
SpecialNewFilesPageObject specialPage = base.openSpecialNewFiles(wikiURL);
```

src/test/java/com/wikia/webdriver/testcases/facebooktests/FacebookButtonTests.java
rochan: you could put this all in one line
```
PreferencesPageObject prefsPage = base.openSpecialPreferencesPage(wikiURL);
```

src/test/java/com/wikia/webdriver/pageobjectsfactory/componentobject/dropdowncomponentobject/DropDownComponentObject.java
jsutterfield: @bekcunning, actually that won't be a problem as long as each subclass defines it's own ```facebookSignUpButton``` instance variable (which you've done). 

So, for example, when you call ```verifyFBButtonVisible()``` on the SpecialUserLoginPageObject [here](https://github.com/Wikia/selenium-tests/blob/UC-190-1/src/test/java/com/wikia/webdriver/testcases/facebooktests/FacebookButtonTests.java#L47), the method inside of base class is called [here](https://github.com/Wikia/selenium-tests/blob/UC-190-1/src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/WikiBasePageObject.java#L1219-L1221), but it'll check if the subclass has a ```facebookSignUpButton``` defined which it does [here](https://github.com/Wikia/selenium-tests/blob/UC-190-1/src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/special/login/SpecialUserLoginPageObject.java#L39-L40). So it'll use the proper selector "#mw-content-text .wikia-button-facebook".

When you call ```verifyFBButtonVisible()``` on the DropDownComponentObject [here](https://github.com/Wikia/selenium-tests/blob/UC-190-1/src/test/java/com/wikia/webdriver/testcases/facebooktests/FacebookButtonTests.java#L29) it'll use [this selector](https://github.com/Wikia/selenium-tests/blob/UC-190-1/src/test/java/com/wikia/webdriver/pageobjectsfactory/componentobject/dropdowncomponentobject/DropDownComponentObject.java#L40-L41) (after you've renamed it to  ```formConnectWithFbButton```).

If the base class doesn't have a ```formConnectWithFbButton``` instance variable defined (like the SpecialNewFilesPageObject class), that's when it'll fall back to the base class version which is what's happening forced login test.

bekcunning: ah, ok, I get it. I might change all the button names to formConnectWithFbButton since that existed before I wrote my tests.

jsutterfield: Sure, up to you. Either is fine as long as we're consistent.

src/test/java/com/wikia/webdriver/testcases/facebooktests/FacebookButtonTests.java
bekcunning: right, sorry didn't see that for some reason.

jsutterfield:
LGTM!

PR #466, opened 2015-01-20
ADEN-1671 Add 71media test params.

drets:


src/test/java/com/wikia/webdriver/common/core/urlbuilder/UrlBuilder.java
```diff
@@ -40,6 +40,9 @@ public String getUrlForWiki(String wikiName) {
 				prefix = "";
 			}
 			suffix = ".com/";
+		} else if (wikiName.endsWith(".de")) {
```
rochan: Do we have any other wikiName that ends with a '.' plus alphabets?
If so, this should be changed to take care of that instead of a particular match for '.de'
e.g. use regular expression / pattern matching

rochan:
LGTM otherwise, the suggestion for the UrlBuilder could always go in as another PR.

src/test/java/com/wikia/webdriver/common/core/urlbuilder/UrlBuilder.java
drets: Thanks for comment. 
I've seen only such situation for German wikis, so if I would see another case I implement it with regex.

PR #467, opened 2015-01-20
Fixing the issue which the parseInt would throw NumberFormatException

rochan:


rochan:
Alternative solution besides the commit, which one is better?
```
	private int getMapValueAsInt(String input) {
		if (input.contains(".")) {
			return (int) Long.parseLong(input);
		} else {
			return Integer.parseInt(input);
		}
	}
```

src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/visualeditor/VisualEditorPageObject.java
```diff
@@ -407,11 +407,19 @@ private Point getBlockTransclusionLocation(int index) {
 		JavascriptExecutor js = (JavascriptExecutor) driver;
 		Object templateBounding = js.executeScript(VEContent.BOUNDING_SCRIPT, blockTransclusionString, index);
 		Map<String, String> mapBounding = (Map) templateBounding;
-		int tempLeft = Integer.parseInt(String.valueOf(mapBounding.get("left")));
```
drets: <code>
int tempLeft = Integer.parseInt(String.valueOf(mapBounding.get("left")));
</code>
to 
<code>
int tempLeft = new BigDecimal(String.valueOf(mapBounding.get("left"))).intValue();
</code>
Does it solve the issue?

rochan: @drets Yup. Thanks for the suggestions.

PR #470, opened 2015-01-21
testing solution with adding a line

justnpT:
test per https://wikia-inc.atlassian.net/browse/QAART-506


rochan:
I have to disagree with this approach. The reason is that the problem is because the dropdown button is not clicked to pull out the drop down menu before clicking on either history/show changes/delete.

justnpT:
Are you sure ? Because my assumption is:
When dropdown appears, any cursor activity outside of the dropdown causes it to disappear.
tested hypothesis: Use mouseover method on dropdown to ensure that cursor doesn't activate outside dropdown area.

rochan:
LGTM. Thanks for the further explanation.

PR #475, opened 2015-01-22
Deleted pomFailed

ludwikkazmierczak:
Deleted pomFailed - the only difference between pomFailed and pom.xml is test suite. Added testSuite property, usage:
```
mvn clean test -DtestSuite=<your_custom_suite.xml>
```

by default - a regular test suite is used


rochan:
LGTM

PR #489, opened 2015-01-29
MON-geo

rochan:
This PR is address two issues are causing MonetizationModuleTest_008-011 to fail.
1) The tests are using the same account, which would cause an early session termination when 1 test is logged in and another test is logging in under the same account.
The solution is to use different accounts for tests that are running in parallel on Jenkins.

2) There is a cached issue with session that even when an user goes to Special:UserLogout page, the session is still logged in when navigate back to the article page. https://wikia-inc.atlassian.net/browse/SOC-233.
The solution is to use Special:UserLogout?returnto=<article> so the test would have a valid anon state on the particular article


saipetch:
LGTM

PR #492, opened 2015-01-30
Google style Fix + Test Fix

tomasznapieralski:
I had big mess in files after pulling master to mercury-002 which has google style formatting. That is why some files were 'edited' but change was only in formatting.
I fixed some mercury tests with FAIL label above method name.


rochan:
LGTM

PR #500, opened 2015-02-09
ADEN-1510 Add providers chain test.

drets:


src/test/java/com/wikia/webdriver/common/templates/NewTestTemplate.java
```diff
@@ -19,11 +21,9 @@ public void beforeClass() {
...
   @BeforeMethod(alwaysRun = true)
   public void start(Method method, Object[] data) {
-    runProxyServerIfNeeded(method);
-    if (method.getAnnotation(UserAgent.class) != null) {
-      setBrowserUserAgent(
-          method.getAnnotation(UserAgent.class).userAgent()
-      );
+    String countryCode = getCountryCode(method, data);
+    if (countryCode != null || method.getAnnotation(NetworkTrafficDump.class) != null) {
+      runProxyServer(countryCode);
```
ludwikkazmierczak: null logic is not very nice - isAnnotationPresent() is a way to go

drets: Yes, it came from old code.

src/test/java/com/wikia/webdriver/common/templates/NewTestTemplate.java
```diff
@@ -43,4 +43,17 @@ public void stop() {
...
+    if (method.getAnnotation(GeoEdgeProxy.class) != null) {
+      return method.getAnnotation(GeoEdgeProxy.class).country();
+    } else {
+      for (Object entity : data) {
+        if (entity instanceof GeoCountry) {
+          return entity.toString();
+        }
+      }
+    }
+    return null;
```
ludwikkazmierczak: Rather than null I would recond using an empty string "" , and .isEmpty() check, also - if you anotate a class, it ignores a data provider, is this a desired behaviour?

src/test/java/com/wikia/webdriver/common/templates/NewTestTemplateCore.java
```diff
@@ -125,41 +122,22 @@ protected void setWindowSize(int width, int height, WebDriver desiredDriver) {
...
-      GeoEdgeUtils geoEdgeUtils = new GeoEdgeUtils(config.getCredentialsFilePath());
-      String credentialsBase64 = "Basic " + geoEdgeUtils.createBaseFromCredentials();
-      String ip = geoEdgeUtils.getIPForCountry(countryCode);
-      networkTrafficIntereceptor.setProxyServer(ip);
-      networkTrafficIntereceptor.changeHeader("Proxy-Authorization", credentialsBase64);
+  public void runProxyServer(String countryCode) {
+    isProxyServerRunning = true;
+    networkTrafficIntereceptor = new NetworkTrafficInterceptor();
+    networkTrafficIntereceptor.startSeleniumProxyServer();
+    if (countryCode != null) {
```
ludwikkazmierczak: another null logic

src/test/java/com/wikia/webdriver/common/templates/NewTestTemplate.java
drets: I think so

src/test/java/com/wikia/webdriver/common/templates/NewTestTemplateCore.java
drets: What's the benefit using empty string against null?

src/test/java/com/wikia/webdriver/common/templates/NewTestTemplate.java
ludwikkazmierczak: my suggestion is to make a method boolean isProxyNeeded() or something like that, insted of checkin nulls and so on.

src/test/java/com/wikia/webdriver/common/templates/NewTestTemplateCore.java
ludwikkazmierczak: no risk of null pointers

ludwikkazmierczak: here is how you can deal with all this null checks = http://en.wikipedia.org/wiki/Null_Object_pattern

drets: Thanks for sharing. Also, http://programmers.stackexchange.com/q/120355/166483

ludwikkazmierczak:
LGTM

PR #501, opened 2015-02-09
QAART-517: Avoided reusing method parameter

rochan:
Method parameters, caught exceptions and foreach variables should not be reassigned

While it is technically correct to assign to parameters from within method bodies, it is better to use temporary variables to store intermediate results. This rule will typically detect cases where a constructor parameter is assigned to itself instead of a field of the same name, i.e. when this was forgotten. Allowing parameters to be assigned to also reduces the code readability as developers will not be able to know whether the original parameter or some temporary variable is being accessed without going through the whole method. Moreover, some developers might also expect assignments of method parameters to be visible from callers, which is not the case and can confuse them. All parameters should be treated as final.


ludwikkazmierczak:
LGTM

PR #508, opened 2015-02-11
Concf 65

justnpT:
Updating Notifications component object for the new notifications


src/test/java/com/wikia/webdriver/pageobjectsfactory/componentobject/notifications/NotificationsComponentObject.java
```diff
@@ -150,6 +149,7 @@ private int getNumberOfUnreadNotifications() {
     for (int i = 0; i < this.notificationsList.size(); i++) {
       WebElement n = this.notificationsList.get(i);
       WebElement nTitle = n.findElement(By.cssSelector(".notification-message h4"));
+      String text = nTitle.getText();
       if (n != null && title.equals(nTitle.getText())) {
```
rochan: Shouldn't this line should use the "text" var from above?

PR #510, opened 2015-02-11
Monetization (MON-348): MonetizationModule - updated tests for checking module width

saipetch:
Monetization (MON-348): update tests for checking module width in MonetizationModule


rochan:
LGTM

PR #513, opened 2015-02-12
finishing test and renaming files for better convention

justnpT:
adding test for verifying that global navigation is attached
renaming classes to assure naming consistency


PiotrGackowski:
LGTM

PR #516, opened 2015-02-13
Monetization (MON-348): MonetizationModule - updated test wiki

saipetch:
Monetization (MON-348): update test wiki and verifyWindowWidth() function

### DIFF
--- a/src/test/java/com/wikia/webdriver/pageobjectsfactory/componentobject/monetizationmodule/MonetizationModuleComponentObject.java
+++ b/src/test/java/com/wikia/webdriver/pageobjectsfactory/componentobject/monetizationmodule/MonetizationModuleComponentObject.java
@@ -190,14 +190,10 @@ public class MonetizationModuleComponentObject extends WikiBasePageObject {
              true);
   }
 
-  public boolean verifyWindowWidth(int width) {
-    if (getWindowSize().width != width) {
-      PageObjectLogging.log("verifyWindowWidth",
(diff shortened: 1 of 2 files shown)


src/test/java/com/wikia/webdriver/pageobjectsfactory/componentobject/monetizationmodule/MonetizationModuleComponentObject.java
```diff
@@ -190,14 +190,10 @@ public void verifyAdsenseUnitNotShownAboveFooter() {
...
-    if (getWindowSize().width != width) {
-      PageObjectLogging.log("verifyWindowWidth",
-                            "The width of the window is invalid (width="
-                            + getWindowSize().width + ")", true);
-      return false;
-    }
-    return true;
+  public boolean verifyWindowWidth(int expectedWidth) {
+    int windowWidth = getWindowSize().width;
+    PageObjectLogging.log("verifyWindowWidth", "The width of the window is " + windowWidth, true);
```
jimmordino: The previous code only logged on error - is there any reason to log if there's no problem?  I'm not a selenium expert, but I've always preferred tests that are emit nothing unless there's a problem because the informational messages make it harder to find the problems.  Do selenium tests typically log informational messages?

saipetch: For debugging

PR #520, opened 2015-02-19
QAART-518

ludwikkazmierczak:
Added a possibility to get a fresh video from youtube using google API v3 search

### DIFF
--- a/src/test/java/com/wikia/webdriver/common/core/CommonExpectedConditions.java
+++ b/src/test/java/com/wikia/webdriver/common/core/CommonExpectedConditions.java
@@ -43,7 +43,7 @@ public class CommonExpectedConditions {
         try {
           String elementsAttributeValue = findElement(locator, from).getAttribute(attribute);
           return elementsAttributeValue.contains(value);
-        } catch (StaleElementReferenceException e) {
+        } catch (StaleElementReferenceException | NullPointerException e) {
           return null;
         }
       }
--- a/src/test/java/com/wikia/webdriver/common/core/CommonExpectedConditions.java
+++ b/src/test/java/com/wikia/webdriver/common/core/CommonExpectedConditions.java
@@ -42,9 +42,13 @@ public class CommonExpectedConditions {
       public Boolean apply(WebDriver from) {
         try {
           String elementsAttributeValue = findElement(locator, from).getAttribute(attribute);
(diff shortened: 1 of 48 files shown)


src/test/java/com/wikia/webdriver/common/core/CommonExpectedConditions.java
```diff
@@ -43,7 +43,7 @@ public Boolean apply(WebDriver from) {
         try {
           String elementsAttributeValue = findElement(locator, from).getAttribute(attribute);
           return elementsAttributeValue.contains(value);
-        } catch (StaleElementReferenceException e) {
+        } catch (StaleElementReferenceException | NullPointerException e) {
```
drets: Looks not nice)

src/test/java/com/wikia/webdriver/common/core/video/Video.java
```diff
@@ -0,0 +1,35 @@
...
+ * Interface used for videos object.
+ */
+
+public interface Video {
+
+  /**
+   * When adding a video to WIKIA, backend is replacing some special characters, make sure you
+   * return a title in the same form that is presented on front-end
+   * 
+   * @return
```
drets: What is the point of return notation?

src/test/java/com/wikia/webdriver/common/core/video/YoutubeVideo.java
```diff
@@ -0,0 +1,61 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Created by Ludwik Kaźmierczak on 2015-02-12.
+ */
+public class YoutubeVideo implements Video {
+
+  static final Map<String, String> TITLE_SPECIAL_CHARS_TO_REPLACE = new HashMap<String, String>() {
```
drets: It would be nice to make this map immutable. You can use ImmutableMap

src/test/java/com/wikia/webdriver/common/core/CommonExpectedConditions.java
ludwikkazmierczak: getAttribute(attribute) returns null if no attribute is found - see get attribute implementation. In that case elementsAttributeValue.contains(value) gives null pointer

drets:
LGTM

PR #524, opened 2015-02-21
SOC-192 test case for facebook account creation without email access

bekcunning:
This test is covering the case where a user signs up for a new account via facebook but doesn't give permission to their email address.

### DIFF
--- a/src/test/java/com/wikia/webdriver/pageobjectsfactory/componentobject/modalwindows/FacebookSignupModalComponentObject.java
+++ b/src/test/java/com/wikia/webdriver/pageobjectsfactory/componentobject/modalwindows/FacebookSignupModalComponentObject.java
@@ -27,7 +27,7 @@ public class FacebookSignupModalComponentObject extends WikiBasePageObject {
   private WebElement createAccountButton;
   @FindBy(css = "#u_0_4")
   private WebElement editInfoProvided;
-  @FindBy(css = "input[type='checkbox'][value='email']")
+  @FindBy(xpath = "//input[@type='checkbox'][@value='email']/..")
   private WebElement emailCheckbox;
 
   String winHandleBefore;
--- a/src/test/java/com/wikia/webdriver/pageobjectsfactory/componentobject/modalwindows/FacebookSignupModalComponentObject.java
+++ b/src/test/java/com/wikia/webdriver/pageobjectsfactory/componentobject/modalwindows/FacebookSignupModalComponentObject.java
@@ -20,8 +20,14 @@ public class FacebookSignupModalComponentObject extends WikiBasePageObject {
   private WebElement usernameField;
   @FindBy(css = ".UserLoginFacebookLeft input[name='password']")
   private WebElement passwordField;
+  @FindBy(css = ".UserLoginFacebookLeft input[name='email']")
+  private WebElement emailField;
   @FindBy(css = ".UserLoginFacebookLeft input[type='submit']")
   private WebElement createAccountButton;
+  @FindBy(css = "#u_0_4")
+  private WebElement editInfoProvided;
+  @FindBy(css = "input[type='checkbox'][value='email']")
+  private WebElement emailCheckbox;
 
   String winHandleBefore;
 
@@ -55,6 +61,29 @@ public class FacebookSignupModalComponentObject extends WikiBasePageObject {
     }
   }
 
+  public void acceptWikiaAppPolicyNoEmail() {
+
+    Set<String> handles = driver.getWindowHandles();
+
+    if (handles.size() > 1) {
+      for (String winHandle : handles) {
+        //Switch to new window opened
+        driver.switchTo().window(winHandle);
+      }
+      waitForElementByElement(editInfoProvided);
+      editInfoProvided.click();
+      PageObjectLogging.log("acceptWikiaAppPolicyNoEmail", "editing info provided", true);
+      emailCheckbox.click();
+      PageObjectLogging.log("acceptWikiaAppPolicyNoEmail", "unchecked the email checkboxbox", true);
+      waitForElementByElement(appTermsConfirmButton);
+      appTermsConfirmButton.click();
+      // Switch back to original browser (first window)
+      driver.switchTo().window(winHandleBefore);
+    } else {
+      PageObjectLogging.log("acceptWikiaAppPolicy", "wikia apps policies already accepted", true);
+    }
+  }
+
   public void typeUserName(String userName) {
     waitForElementByElement(usernameField);
     usernameField.sendKeys(userName);
@@ -65,6 +94,12 @@ public class FacebookSignupModalComponentObject extends WikiBasePageObject {
     waitForElementByElement(passwordField);
     passwordField.sendKeys(password);
     PageObjectLogging.log("typePassword", "password typed into the field", true);
+  }
+
+  public void typeEmail(String email) {
+    waitForElementByElement(emailField);
+    emailField.sendKeys(email);
+    PageObjectLogging.log("typeEmail", "email typed into the field", true);
   }
 
   public void createAccount() {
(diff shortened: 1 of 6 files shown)


src/test/java/com/wikia/webdriver/testcases/facebooktests/FacebookTests.java
```diff
@@ -29,4 +36,47 @@ public void Facebook_001_removeWikiaApps() {
     settingsFB.openApps();
     settingsFB.removeAppIfPresent();
   }
+
+  @Test(groups = {"Facebook_002", "Facebook"})
+  public void Facebook_002_noEmailPerms() {
```
justnpT: This test is too long, it should be separated into dependent parts

src/test/java/com/wikia/webdriver/testcases/facebooktests/FacebookTests.java
```diff
@@ -29,4 +36,47 @@ public void Facebook_001_removeWikiaApps() {
...
+    String emailPassword = credentials.emailPasswordQaart2;
+    MailFunctions.deleteAllEmails(email, emailPassword);
+    fbModal.typeUserName(userName);
+    fbModal.typePassword(password);
+    fbModal.typeEmail(email);
+    fbModal.createAccount();
+    signUp.verifyUserLoggedIn(userName);
+    signUp.verifyEmailNotConfirmed();
+    try {
+      Thread.sleep(5000);
```
justnpT: what is the reason for using arbitrary sleep, is any wait method not aplicable here ?

src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/facebook/FacebookSettingsPageObject.java
```diff
@@ -64,14 +66,19 @@ public void removeAppIfPresent() {
...
     }
-    PageObjectLogging.log("removeApp", "Wikia App removed", true);
   }
 
   /**
    * This method verifies if App is present on facebook apps list. It searches for Wikia string.
    */
   private boolean isAppPresent() {
+    try {
+      Thread.sleep(3000);
```
justnpT: what is the reason for using arbitrary sleep, is any wait method not aplicable here ?

justnpT:
I tried to run this test and it never passed for me with success
The reason is that after it logged in to facebok, clickin 'sign up with facebook' button on Special:Signup triggers facebook .login. so no Wikia Policy permision window is opened.

The logic of the test should be once again thought so that it handles all the specific conditions.
The test should be able to pass always, test stability is highly wanted
If necessary new facebook credentials should be created with dedicated purpose for this test

justnpT:
please don't merge it before the test gets its stability.

bekcunning:
@justnpT the test requires that you run the previous test case first Facebook_001_removeWikiaApps() - which will delete the app from the facebook account and thus the permission dialog will appear when you run this test.  Is there a way that I should indicate that in the test structure?  Still new to this process and figuring it out as I go.

src/test/java/com/wikia/webdriver/testcases/facebooktests/FacebookTests.java
bekcunning: I used sleep because I wanted to give time for the email to arrive in the inbox, as I was having timing issues with trying to find the email before it arrived.  If you know a better solution I am happy to hear it.

src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/facebook/FacebookSettingsPageObject.java
bekcunning: The facebook/settings applications page takes a few seconds to load the Apps list -- I cannot wait for a specific element, as I don't know ahead of time if it will be on the page for certain, so I'd rather wait a determined amount of time before checking. Waiting for the high level element ID list was causing a failure regularly and this helps minimize that failure case.  If you have other ideas I'm happy to hear them.

justnpT:
You can add dependency method to the test annotation.
In this case it should look like this:
 @Test(groups = {"Facebook_002", "Facebook"},
      dependsOnMethods = {"Facebook_001_removeWikiaApps"})

src/test/java/com/wikia/webdriver/testcases/facebooktests/FacebookTests.java
justnpT: We have an already implemented method for waiting for an email to arrive.
See: 
package com.wikia.webdriver.common.core; 
class: MailFunctions
method: getFirstEmailContent

The logic in current method is assuming that the arriving email is a password reminder. This is not well generalized. You can create more general method using the same logic, which just waits for new e-mail to arrive. The code is allready there

src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/facebook/FacebookSettingsPageObject.java
justnpT: Ok, if there really is no such determining element (that will be there for certain), then it is ok. Just make sure the 3000 ms time is the best choice. As you can see we wan't to minimize false negatives as much as possible, so that the tests are stable and reliable. And the set sleep periods often create a risk for the stability of a test.

src/test/java/com/wikia/webdriver/testcases/facebooktests/FacebookTests.java
justnpT: ForgottenPassword test is the only test that uses this method, you can take a look how it does it if you want, because its intuitive anyway :)

bekcunning:
Ok, I've tidied it up a bit.  I couldn't see a way to split it into multiple test cases since I need to use the same user instance throughout.  If you have ideas of how to do that let me know.  I pushed some of the functionality into the page objects which required sending many parameters -- I don't love using so many parameters, but I think hopefully it cleans it up so that it is a little more readable.

src/test/java/com/wikia/webdriver/testcases/facebooktests/FacebookTests.java
```diff
@@ -29,4 +33,45 @@ public void Facebook_001_removeWikiaApps() {
...
+    String emailPassword = credentials.emailPasswordQaart2;
+    fbModal.createAccountNoEmail(email, emailPassword, userName, password);
+    signUp.verifyUserLoggedIn(userName);
+
+    AlmostTherePageObject almostThere = new AlmostTherePageObject(driver);
+    almostThere.confirmAccountAndLogin(email, emailPassword, userName, password, wikiURL);
+    almostThere.logOut(driver);
+
+    SignUpPageObject signUp2 = userFB.openSpecialSignUpPage(wikiURL);
+    FacebookSignupModalComponentObject fbModal2 = signUp2.clickFacebookSignUp();
```
justnpT: this field is not used

src/test/java/com/wikia/webdriver/pageobjectsfactory/componentobject/modalwindows/FacebookSignupModalComponentObject.java
```diff
@@ -55,6 +62,29 @@ public void acceptWikiaAppPolicy() {
...
+
+    if (handles.size() > 1) {
+      for (String winHandle : handles) {
+        //Switch to new window opened
+        driver.switchTo().window(winHandle);
+      }
+      waitForElementByElement(editInfoProvided);
+      editInfoProvided.click();
+      PageObjectLogging.log("acceptWikiaAppPolicyNoEmail", "editing info provided", true);
+      emailCheckbox.click();
```
justnpT: i'm getting element not clickable error at this point.
I don't see checkbox on the page when the test execution reaches this line, and it fails
We really don't want fresh tests to fail ! :)

justnpT: Thank you for the fixes,
It tried to run the test and It still doesn't pass successfully for me, I put comment where it fails.
Is it always passing for you ? It has to always pass otherwise it's just unstable

bekcunning:
@justnpT did the previous test pass successfully?  Can you check the facebook account to see if the wikia app was actually removed?  As long as the wikia app gets removed by the dependent case, then I see this test case pass 100% for me.  This would be a lot easier to figure out if we were in the same office. :-\

PR #526, opened 2015-02-23
ADEN-1743 Change functionality for hiding slot to changing opacity property of the web element.

drets:
We moving 'hide/show slot' functionality from changing visibility to changing opacity because we found one case when changing visibility of slot doesn't hide the ad (if between slot and ad we have items with visibility equals show).


rochan:
LGTM

PR #530, opened 2015-02-25
Wikiamobile tests fix

tomasznapieralski:
Because you can obtain mercury skin with 50% chance i added beforeMethod where i log in as qatestuser to avoid obtaining mercury skin


RodriGomez:
LGTM

PR #531, opened 2015-02-25
removing photos module check, reorganising tests to proper order

justnpT:
Removing photos module from right rail
selenium tests update

https://wikia-inc.atlassian.net/browse/CONCF-116


ludwikkazmierczak:
LGTM

PR #532, opened 2015-02-26
Small interactive maps tests for wikiamobile skin

tomasznapieralski:
It's only for two weeks till mercury will be in 100%.
It's small tests and i decided to not separate it to two files, that is why i use Thred.sleep and locating elements inside tests etc.


RodriGomez:
@ludwikkazmierczak the reason behind no pageobject class or storing constant values is that this 2 tests will be deleted in a week or two, once Mercury is enabled in 100% of traffic. We just need something quick for the next days. The PR LGTM but Im leaving it up to you to merge.

PR #549, opened 2015-03-06
Fixed facebook_003 test

tomasznapieralski:
Problems:
- When clicking on facebook button via selenium dropdown disappears --> changed to js click (Bogna's idea)
- Double call method Facebook_001_removeWikiaApps --> removed from Facebook_003_linkFBwithWikia

Applied code format for DropDownComponentObject.java and FacebookTests.java


src/test/java/com/wikia/webdriver/pageobjectsfactory/componentobject/dropdowncomponentobject/DropDownComponentObject.java
```diff
@@ -133,7 +132,10 @@ public void fillPasswordInput(String password) {
   }
 
   public void logInViaFacebook(String email, String password) {
-    scrollAndClick(formConnectWithFbButton);
```
justnpT: what's the purpose of that change ?

tomasznapieralski: Yes it does.

PR #561, opened 2015-03-11
removig deleted reintroduces search 14 test, changing result string

justnpT:
1 . searchPageOpened  was deleted earlier this week, but a merge reintroduced it. 
2. anonSearch and userSearch tests are searching were searching for lowerCase letter a, when result was upperCase letter A

PR provides fix for both
1. removes searchPageOpened test
2. changes search query from lower-case to upper-case


RodriGomez:
LGTM

PR #573, opened 2015-03-19
Selenium 2450

ludwikkazmierczak:
Updated dependencies. Apart form selenium I needed to upgrade gson dependency. Long story short - gson is a dependency used by selenium and we were overriding that version, causing json parser to work in an old version.

I've tested on a full matrix run - no tests broken cause of updated libs.


drets:
LGTM :+1:

PR #576, opened 2015-03-20
ADEN-1846 Update krux tests.

drets:


src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/adsbase/AdsKruxObject.java
```diff
@@ -49,17 +46,23 @@ public void verifyKruxControlTag(String kruxSiteId) {
    * Test whether the Krux user id is not empty and added to GPT calls
    */
   public void verifyKruxUserParam() {
+    String script = "return localStorage.kxuser;";
+    waitForKrux();
+    String user1 = (String) ((JavascriptExecutor) driver).executeScript(script);
```
nandy-andy: How about adding a `TODO:` comment? We should discuss it and if possible find out why it happens...

nandy-andy:
LGTM :+1: It'd be awesome if we could get rid off the `user1`, `user2` hack but maybe it's just how Krux works with `WebDriver` :/

PR #577, opened 2015-03-20
VE-1874 Updating test for media setting

rochan:
UI mapping update for media settings dialog. Custom size now has two input fields. Updated the test to reflect that.
UI mapping update for context menu

### DIFF
--- a/src/test/java/com/wikia/webdriver/common/dataprovider/VisualEditorDataProvider.java
+++ b/src/test/java/com/wikia/webdriver/common/dataprovider/VisualEditorDataProvider.java
@@ -92,6 +92,11 @@ public class VisualEditorDataProvider {
     ADVANCED
   }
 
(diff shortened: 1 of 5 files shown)


src/test/java/com/wikia/webdriver/common/dataprovider/VisualEditorDataProvider.java
```diff
@@ -92,6 +92,11 @@ public By getTag() {
     ADVANCED
   }
 
+  public enum ImageSize {
```
RodriGomez: For my interest Robert, why do these have to be enum?

rochan: Enums are lists of constants. When you need a predefined list of values which do not represent some kind of numeric or textual data, you should use an enum.

PR #579, opened 2015-03-23
VE-1878: Updating test for adding and editing templates

rochan:
Added wait for edit template's focus mode for a better transition.
Refactored the method for transclusion to be generic.
Removed the unnecessary try and catch since the test no longer needs to switch back to default content.

### DIFF
--- a/src/test/java/com/wikia/webdriver/common/dataprovider/VisualEditorDataProvider.java
+++ b/src/test/java/com/wikia/webdriver/common/dataprovider/VisualEditorDataProvider.java
@@ -92,6 +92,21 @@ public class VisualEditorDataProvider {
     ADVANCED
   }
 
+  public enum Transclusion {
+    INLINE("span[typeof='mw:Transclusion']"),
+    BLOCKED("div[typeof='mw:Transclusion']");
+
+    private String cssSelector;
+
+    private Transclusion(String cssSelector) {
+      this.cssSelector = cssSelector;
+    }
+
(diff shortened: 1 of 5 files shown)


src/test/java/com/wikia/webdriver/common/dataprovider/VisualEditorDataProvider.java
```diff
@@ -92,6 +92,21 @@ public By getTag() {
     ADVANCED
   }
 
+  public enum Transclusion {
```
ludwikkazmierczak: I would move it to another class, data provider seems to be a bad place for such a nested class

src/test/java/com/wikia/webdriver/pageobjectsfactory/componentobject/visualeditordialogs/VisualEditorInsertTemplateDialog.java
```diff
@@ -44,71 +49,46 @@ public void typeInSearchInput(String searchString) {
...
         true
     );
-    switchOutOfIFrame();
   }
 
   public void clearSearchInput() {
-    switchToIFrame();
-    waitForElementVisibleByElement(searchClearButton);
-    searchClearButton.click();
+    searchInput.clear();
```
ludwikkazmierczak: sure you dont wanna wait for this button being clickable?

ludwikkazmierczak:
LGTM

PR #580, opened 2015-03-24
adding global navigation tests to suite

justnpT:


justnpT:
adding global navigation tests to suite
updating group name for scrollDown test

ludwikkazmierczak:
LGTM

PR #583, opened 2015-03-25
VE-1880 Updated the test for link inspector

rochan:
Updated the UI mapping and logic for the suggestion since now it's very easy to identify what's the suggestion category vs suggestion result. Now there are less work to determine what's what


ludwikkazmierczak:
LGTM

PR #596, opened 2015-04-02
Adding coverage for connecting FB account through prefs

bekcunning:


rochan:
Try to run the group and/or method as a whole with maven command to make sure that there's no surefire dependency issues.

ludwikkazmierczak:
LGTM

PR #598, opened 2015-04-03
Facebook tests - fix for QAART-559

bekcunning:
moved the remove facebook function to it's own page object and added an after method to clean up prefs when test is finished running.


ludwikkazmierczak:
LGTM

PR #602, opened 2015-04-07
QAART-537 Fix for RTE_019_MoreSympolsTools

rochan:
The test would fail sometimes if the test tries to click on the submit button when the modal is up. When the modal is up, is when there is a white overlay that is blocking the button. Adding wait for those items to be gone after clicking the button to close the modal.


RodriGomez:
LGTM

PR #604, opened 2015-04-07
QAART-521: Fixing the ImageStorage001 test

rochan:
The issue with the test failing is because the selector is not targeting the latest file that was uploaded. Since each individual uploaded file has a unique random filename, sometimes when it would select a different file that the value of 'src' would not match.

### DIFF
--- a/src/test/java/com/wikia/webdriver/testcases/imageservingtests/ImageStorageTests.java
+++ b/src/test/java/com/wikia/webdriver/testcases/imageservingtests/ImageStorageTests.java
@@ -3,9 +3,6 @@
  */
 package com.wikia.webdriver.testcases.imageservingtests;
 
-import org.joda.time.DateTime;
-import org.testng.annotations.Test;
-
 import com.wikia.webdriver.common.contentpatterns.PageContent;
 import com.wikia.webdriver.common.driverprovider.UseUnstablePageLoadStrategy;
 import com.wikia.webdriver.common.properties.Credentials;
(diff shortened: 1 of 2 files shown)


src/test/java/com/wikia/webdriver/testcases/imageservingtests/ImageStorageTests.java
```diff
@@ -32,9 +32,9 @@
 
   @Test(groups = {"ImageStorageTests", "ImageStorage_001"})
   @UseUnstablePageLoadStrategy
-  public void ImageStorage_001_deleteImage_QAART_521() {
+  public void ImageStorage_001_deleteImage() {
     WikiBasePageObject base = new WikiBasePageObject(driver);
-    base.logInCookie(credentials.userNameStaff, credentials.passwordStaff, wikiURL);
+    base.logInCookie(credentials.userNameStaff2, credentials.passwordStaff2, wikiURL);
```
ludwikkazmierczak: Why use Staff2 insted of staff?

rochan: I also saw an error on this test that when the test gets to the part where it needs to delete the file, the page shows that the user doesn't have permission. In fact, the log shows the user is no longer logged in. 
This is just an attempt to see if other tests are causing the user to get booted out of the session

ludwikkazmierczak:
LGTM

PR #605, opened 2015-04-07
wait for signup dialog to go away before checking if user is logged in

bekcunning:


rochan:
Would suggest using navigateToSpecialSignUpPage instead of openSpecialSignUpPage to bypass the nav bar + UI elements

PR #607, opened 2015-04-08
MON-380 Updated the test wiki and article for ecommerce multi product li...

rochan:
...ght skin

https://wikia-inc.atlassian.net/browse/MON-380


saipetch:
LGTM

PR #611, opened 2015-04-10
QAART-568: Adding a parameter to disable Adobe flash (for FF)

rochan:
https://wikia-inc.atlassian.net/browse/QAART-568
Disabling Adobe Flash to ensure that FF would exit all of the plugins before the driver exits.
This would make the FF run time to be normal again as org.openqa.selenium.os.UnixProcess$SeleniumWatchDog destroyHarder should no longer happen on Jenkins

### DIFF
--- a/src/test/java/com/wikia/webdriver/common/core/configuration/ManualConfiguration.java
+++ b/src/test/java/com/wikia/webdriver/common/core/configuration/ManualConfiguration.java
@@ -90,4 +90,9 @@ public class ManualConfiguration extends AbstractConfiguration {
   public String getDeviceName() {
     return config.get("device-name");
   }
+
+  @Override
+  public String getDisableFlash() {
+    return config.get("disable-flash");
+  }
 }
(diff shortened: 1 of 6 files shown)


src/test/java/com/wikia/webdriver/common/core/configuration/ManualConfiguration.java
```diff
@@ -90,4 +90,9 @@ public String getAppiumIp() {
   public String getDeviceName() {
     return config.get("device-name");
   }
+
+  @Override
+  public String getDisableFlash() {
+    return config.get("disable-flash");
```
ludwikkazmierczak: if you put in config file:
```disable-flash: true```
this wont work - can't cast boolean to String.
A solution:
```String.valueOf(config.get("disable-flash"));```

File: src/test/java/com/wikia/webdriver/common/driverprovider/NewDriverProvider.java [hunk 2]
```diff
@@ -199,6 +199,10 @@ private static EventFiringWebDriver getFFInstance() {
       firefoxProfile.setPreference("webdriver.load.strategy", "unstable");
     }
 
+    if (StringUtils.isNotBlank(ConfigurationFactory.getConfig().getDisableFlash().toString())) {
+      firefoxProfile.setPreference("plugin.state.flash", 0);
+    }
```
ludwikkazmierczak: ```.toString() ```
is not needed here

PR #614, opened 2015-04-10
QAART-565 Removed incorrect double logging in assertion

rochan:
https://wikia-inc.atlassian.net/browse/QAART-565
Previously if a test fails, it would log the failure then also log the success, this creates double logging and incorrect logging as well.
Also formatted the log output to put double quotes on the parameter passed into the assertion.


ludwikkazmierczak:
LGTM

PR #616, opened 2015-04-13
Mercury tests improvement

tomasznapieralski:
Changes summary:
- combined tests: 57 -> 27;
- decreased build execution time: 55 min -> 21 min;
- increased build stability: 80% -> 99%
- removed "opt in mercury skin" step;
- decreased timeout for specific tests: 30 sec -> 5 sec;
- replaced static wait for reloaded page by watching reload spinner;
- used combination of assertion and page object logging to get more information in log;
- added fail control by setting flag in methods;
- unified mercury tests structure;
- changed methods names;


RodriGomez:
@tomasznapieralski did this PR receive any CR?

src/test/java/com/wikia/webdriver/common/logging/PageObjectLogging.java
```diff
@@ -306,4 +307,8 @@ private void appendShowHideButtons() {
   public void onFinish(ITestContext context) {
     CommonUtils.appendTextToFile(logPath, "</body></html>");
   }
+
+  public static void appendTextToLogFile(StringBuilder builder) {
+    CommonUtils.appendTextToFile(logPath, builder.toString());
```
ludwikkazmierczak: Please add javadoc - what this text is supposed to do?

src/test/java/com/wikia/webdriver/pageobjectsfactory/componentobject/mercury/InteractiveMapsComponentObject.java
```diff
@@ -127,20 +111,12 @@ public boolean isZoomInButtonEnabled() throws WebDriverException {
...
-    return !ic.areFilesTheSame(beforeScrolling, afterScrolling);
-  }
-
   public void switchToMapFrame() {
     waitForElementByElement(lightbox);
     driver.switchTo().frame(mapFrame);
   }
+
+  public void switchToDefaultFrame() {
+    driver.switchTo().defaultContent();
```
ludwikkazmierczak: This method is used just once, and it seems like it's just noise - in my opinion, it's not needed and can be deleted. Is there any reason not to use ```driver.switchTo().defaultContent();``` in place this method is invoked?

src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/mercury/BasePageObject.java
```diff
@@ -43,17 +46,44 @@ public void tapOnElement(WebElement element) {
...
+
+  /**
+   * First waits for spinner to be visible and then waits for spinner to be hidden
+   */
+  public void waitForLoadingSpinnerToFinishReloadingPage() {
+    try {
+      waitForElementByElement(loadingSpinner);
+      waitForElementPresenceByBy(By.cssSelector(".loading-overlay.hidden"));
+    } catch (TimeoutException e) {
+    }
```
ludwikkazmierczak: Why this catch is empty? Is situation when spinner is not hidden ok?

src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/mercury/BasePageObject.java
```diff
@@ -43,17 +46,44 @@ public void tapOnElement(WebElement element) {
...
+  }
+
+  /**
+   * Example: wait 10 sec for element and check each 0.5 for that element (10 / 0.5 = 20 attempts)
+   *
+   * @param element            WebElement
+   * @param timeOutInSec       int
+   * @param checkOutInMilliSec int
+   */
+  public void waitForElementVisibleByElementCustomTimeOut(WebElement element, int timeOutInSec,
```
ludwikkazmierczak: Is CustomTimeout postfix need here? Seems like overloading ```waitForElementVisibleByElement``` is a good idea.

src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/mercury/BasePageObject.java
```diff
@@ -43,17 +46,44 @@ public void tapOnElement(WebElement element) {
...
+    } finally {
+      restoreDeaultImplicitWait();
     }
   }
 
-  public static void turnOnMercurySkin(WebDriver driver, String wikiURL) {
-    BasePageObject base = new BasePageObject(driver);
-    SpecialMercuryPageObject mercuryPage = base.openSpecialMercury(wikiURL);
-    mercuryPage.clickMercuryButton();
+  public void failTest(boolean fail) {
```
ludwikkazmierczak: Could you please add JavaDoc? Why you need this method?

src/test/java/com/wikia/webdriver/testcases/mercurytests/InteractiveMapsTests.java
```diff
@@ -59,77 +79,68 @@ public void MercuryInteractiveMaps_003_ZoomByGesture() {
...
+    afterZooming = new Shooter().capturePage(driver);
+    Assertion.assertFalse(new ImageComparison().areFilesTheSame(beforeZooming, afterZooming),
+                          "Zoom out doesn't work");
+    PageObjectLogging.log("Zoom out by gesture", "works", true);
+    Assertion.assertTrue(maps.isZoomInButtonEnabled(), "Zoom in button is disabled");
+    PageObjectLogging.log("Zoom in button", "is enabled", true);
+    beforeZooming = new Shooter().capturePage(driver);
+    touchAction.zoomInOutPointXY(50, 50, 50, 100, PerformTouchAction.ZOOM_WAY_IN, 5000);
+    afterZooming = new Shooter().capturePage(driver);
+    Assertion.assertFalse(new ImageComparison().areFilesTheSame(beforeZooming, afterZooming),
```
ludwikkazmierczak: In some cases - if zoom is broken (Let's say instead of zooming, you will move a view by 1px), test will finish with positive result. Maybe consider using something like this to check if page is zoomed ```$(window).width()/screen.width```
Probably you should add some pixel ratio scaling for mobile devices, but it's worth trying , should be a lot faster then comparing screenshots

src/test/java/com/wikia/webdriver/testcases/mercurytests/LightboxTests.java
```diff
@@ -40,61 +38,154 @@ public void optInMercury() {
...
+    if (new ImageComparison().areFilesTheSame(beforeZooming, afterZooming)) {
+      PageObjectLogging.log("Zooming in by double tap", "does not work", false);
+      failTest = true;
+    } else {
+      PageObjectLogging.log("Zooming in by double tap", "works", true);
+    }
+    touchAction.tapOnPointXY(50, 50, 140, 0);
+    touchAction.tapOnPointXY(50, 50, 140, 3000);
+    afterZooming = new Shooter().capturePage(driver);
+    if (new ImageComparison().areFilesTheSame(beforeZooming, afterZooming)) {
```
ludwikkazmierczak: Same as before - consider using something else then image comparison

src/test/java/com/wikia/webdriver/pageobjectsfactory/componentobject/mercury/InteractiveMapsComponentObject.java
tomasznapieralski: I need it, because without that selenium couldn't find webelement

src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/mercury/BasePageObject.java
tomasznapieralski: Spinner must be hidden after page loads

src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/mercury/BasePageObject.java
tomasznapieralski: I need that in situation, when i had element and then it was deleted from DOM, so i want to check that this element is no longer present

src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/mercury/BasePageObject.java
tomasznapieralski: Because if i mark test as fail from pageobjectlogging, it still mark tests as success in jenkins. But i want to mark failed functionality and if i can continue executing test and at the end i want to fail test

src/test/java/com/wikia/webdriver/testcases/mercurytests/InteractiveMapsTests.java
tomasznapieralski: Probably you right, but this PR doesn't add new functionality, this code comes from other method which did too much like isZoomWorks, in the future we change that.

src/test/java/com/wikia/webdriver/testcases/mercurytests/LightboxTests.java
tomasznapieralski: Answer above

src/test/java/com/wikia/webdriver/common/logging/PageObjectLogging.java
tomasznapieralski: logPath is private, that is why i add method to add custom log to it, in my case i would add warning class to log line

src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/mercury/BasePageObject.java
```diff
@@ -15,17 +22,13 @@
  */
 public class BasePageObject extends MobileBasePageObject {
```
justnpT: I would suggest taking a step back with this architecture:
This class is called BasePageObject. It's risky. 
There is already a BasePageObject class which is considered to be wikia-independent. So it doesn't contain any wikia-related stuff, and that is its purpose.

So the mercury/BasePageObject:
- is a false duplicate, doesn't follow the wikia-independent purpose 
- extends MobileBasePageObject which makes it less "base", when its name suggests being more "base"

justnpT:
On high level:
I don't like keeping everything in pageobjectsfactory/pageobject/mercury/ folder
In the end mercury is in fact "mobile", and will soon be 100% of our mobile traffic
We should change the naming in the future so that it can be easily understood by all.

Mercury is the risky type of project naming, where you name a lot of things under one name that doesn't have anything to do with what it touches

src/test/java/com/wikia/webdriver/pageobjectsfactory/componentobject/mercury/InteractiveMapsComponentObject.java
ludwikkazmierczak: Question is - why you need ```switchToDefaultFrame``` method, can't we just use```driver.switchTo().defaultContent();``` where it's needed?

tomasznapieralski:
When wikiamobile test won't be necessary we can change naming

src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/mercury/BasePageObject.java
ludwikkazmierczak: in this case, if spinner is visible, and then stays on page this method just goes on after 30 sec.

src/test/java/com/wikia/webdriver/testcases/mercurytests/ArticlePageTests.java
```diff
@@ -18,88 +19,113 @@
...
   private static final String[]
       FOOTER_ELEMENTS =
       {"Games", "Movies", "TV", "Comics", "Music", "Books", "Lifestyle", "Full site", "Licensing",
        "Privacy Policy", "Feedback"};
 
   // APT01
-  @Test(groups = {"MercuryArticleTests_001", "MercuryArticleTests", "Mercury"})
-  public void MercuryArticleTests_001_LogoAndSearchButtonAreVisible() {
+  @Test(groups = {"MercuryArticleTest_001", "MercuryArticleTests", "Mercury"})
+  public void MercuryArticleTest_001_Logo_Search_TopContributors_FooterElements() {
```
justnpT: Use block of assertion instead of  if / else
Then you can remove the base.failTest(failTest); thing

Assertion.assertTrue(articlePage.isSearchButtonVisible())

src/test/java/com/wikia/webdriver/testcases/mercurytests/CommentsTests.java
```diff
@@ -18,39 +19,63 @@
...
   private static final String MEDIA_TYPE_VIDEO = "Video";
   private static final String MEDIA_TYPE_IMAGE = "Image";
   private static final int NUMBER_OF_COMMENTS_PER_PAGE = 25;
 
   // CT01
-  @Test(groups = {"MercuryCommentsTests_001", "MercuryCommentsTests", "Mercury"})
-  public void MercuryCommentsTests_001_ClickingCommentsWillUncollapseComments() {
+  @Test(groups = {"MercuryCommentsTest_001", "MercuryCommentsTests", "Mercury"})
+  public void MercuryCommentsTest_001_CommentsList_Avatar_Username_TimeStamp_Content() {
     BasePageObject base = new BasePageObject(driver);
```
justnpT: Use block of assetions

src/test/java/com/wikia/webdriver/testcases/mercurytests/ArticlePageTests.java
tomasznapieralski: I used if..else because i don't want to stop test, i want to continue it and check other things

src/test/java/com/wikia/webdriver/testcases/mercurytests/CommentsTests.java
```diff
@@ -60,106 +85,104 @@ public void MercuryCommentsTests_003_NextAndPreviousPageAreVisible() {
...
     comments.clickCommentsHeader();
-    Assertion.assertTrue(
-        comments.getNumberOfRepliesFromHeader(0) == comments.getNumberOfRepliesFromList(0),
-        "Replies counter doesn't work");
+    Assertion.assertFalse(comments.isRepliesListExpanded(), "Replies list is expanded");
+    PageObjectLogging.log("Replies list", "is collapsed", true);
+    comments.clickViewReplies(0);
+    Assertion.assertTrue(comments.isRepliesListExpanded(), "Replies list is collapsed");
+    PageObjectLogging.log("Replies list", "is expanded", true);
+    if (comments.getNumberOfRepliesFromHeader(0) == comments.getNumberOfRepliesFromList(0)) {
```
justnpT: use assertion here
headerReplies = comments.getNumberOfRepliesFromHeader(0)
listReplies = comments.getNumberOfRepliesFromList(0)
Assertion.asserEquals(headerReplies, listReplies)

justnpT:
Generally this logic is often use:
 if (condition) {log pass} else {failTest = true} :
It should be replaced with assertion
Assertion.assert$coparer(condition)

tomasznapieralski:
I used if..else because i don't want to stop test, i want to continue it and check other things.
But i will compress that.

tomasznapieralski:
Fix is here #617

PR #617, opened 2015-04-14
Fix to PR #616

tomasznapieralski:
Changes summary:
- removed fail check <-- [QAART-576](https://github.com/Wikia/selenium-tests/tree/QAART-576) fix that problem (already included);
- fixed unhandled exception from loading spinner;
- added PageObjectLogging::logWarning() method, which adds yellow line to log;
- added PageObjectLogging::log() method, which will log different message on success and fail to log;
- added BasePageObject::waitForElementVisibleByElement() method, which will wait declared time;
- replaced logging with if..else to new log() method;

### DIFF
--- a/src/test/java/com/wikia/webdriver/common/logging/PageObjectLogging.java
+++ b/src/test/java/com/wikia/webdriver/common/logging/PageObjectLogging.java
@@ -79,6 +79,20 @@ public class PageObjectLogging extends AbstractWebDriverEventListener implements
     log(command, description, success, false);
   }
 
+  /**
+   * This method will log warning to log file (line in yellow color)
+   *
+   * @param command
+   * @param description
+   */
+  public static void logWarning(String command, String description) {
+    StringBuilder builder = new StringBuilder().append("<tr class=\"warning\">"
+                                                       + "<td>" + command + "</td>"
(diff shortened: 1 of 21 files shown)


tomasznapieralski:
**DON'T MERGE BEFORE QAART-576**

rochan:
@tomasznapieralski  @ludwikkazmierczak  Is this https://github.com/Wikia/selenium-tests/commit/8072ec3d8394339f5008a2c9c7a17663cf8729cc an intentional commit?

src/test/java/com/wikia/webdriver/common/core/exceptions/TestFailed.java
```diff
@@ -0,0 +1,12 @@
+package com.wikia.webdriver.common.core.exceptions;
+
+/**
+ * Created by Ludwik on 2015-04-14.
+ */
+public class TestFailed extends Throwable {
```
rochan: Question: Is there something specific of this class that it needs to extend Throwable and not Exception?

rochan: Class name of this should reflect that it is a throwable or exception

tomasznapieralski:
Fix is here #621

PR #620, opened 2015-04-15
Qaart 576

ludwikkazmierczak:
Now, any test step marked by log as failed with:
```PageObjectLogging.log("My command", "My super message", false);```
will fail whole test executed


tomasznapieralski:
LGTM, works on Jenkins and IntelliJ

PR #621, opened 2015-04-15
Fix to PR #616

tomasznapieralski:
Changes summary:
- removed fail check <-- QAART-576 fix that problem (already included);
- fixed unhandled exception from loading spinner;
- added PageObjectLogging::logWarning() method, which adds yellow line to log;
- added PageObjectLogging::log() method, which will log different message on success and fail to log;
- added BasePageObject::waitForElementVisibleByElement() method, which will wait declared time;
- replaced logging with if..else to new log() method;


ludwikkazmierczak:
LGTM

PR #622, opened 2015-04-16
[Mobile Apps] Test ponto request in interactive maps service

Dyktus:
This test check get response from ponto. We use this response in mobile community apps.

### DIFF
--- a/src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/special/interactivemaps/InteractiveMapPageObject.java
+++ b/src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/special/interactivemaps/InteractiveMapPageObject.java
@@ -168,7 +168,7 @@ public class InteractiveMapPageObject extends BasePageObject {
   }
 
   public void clickOnPin(Integer pinListPosition, boolean ...noFrame) {
-    if(!noFrame[0]) {
+    if(noFrame.length == 0) {
         waitForElementVisibleByElement(mapFrame);
         driver.switchTo().frame(mapFrame);
     }
@@ -234,7 +234,7 @@ public class InteractiveMapPageObject extends BasePageObject {
   }
 
   public void clickOpenPinTitle(boolean ...noFrame) {
-    if(!noFrame[0]) {
+    if(noFrame.length == 0) {
         waitForElementVisibleByElement(mapFrame);
         driver.switchTo().frame(mapFrame);
     }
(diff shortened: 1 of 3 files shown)


src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/special/interactivemaps/InteractiveMapPageObject.java
```diff
@@ -166,9 +167,11 @@ public void clickZoomOutButton() {
     PageObjectLogging.log("clickZoomOutButton", "Map zoom out was clicked", true, driver);
   }
 
-  public void clickOnPin(Integer pinListPosition) {
-    waitForElementVisibleByElement(mapFrame);
-    driver.switchTo().frame(mapFrame);
+  public void clickOnPin(Integer pinListPosition, boolean ...noFrame) {
```
RodriGomez: Can you share why this boolean parameter is needed?

src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/special/interactivemaps/InteractiveMapPageObject.java
```diff
@@ -503,4 +508,12 @@ public void verifyPoiPointDescription() {
     PageObjectLogging.log("verifyPoiPointDescription",
                           "Poi description section is displayed", true);
   }
+
+  public void verifyPontoGetRequest(NetworkTrafficInterceptor networkTab) {
+    if(networkTab.searchRequestUrlInHar("maps.wikia-services.com")) {
```
RodriGomez: Basically, the test wants to check from Network tab if there is a request to maps.wikia-services.com right? Is this enough to assure the request from Ponto came?

src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/special/interactivemaps/InteractiveMapPageObject.java
Dyktus: Because if I'm going to 
maps.wikia-services.com/api/v1/render/77/?hideAttr=1&isEmbeddedInWikiaApp&uselang=en
There is no iframe and I don't want to switch there. I could overwrite this method and skip driver.switchTo() but I prefered to add this optional method argument.

src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/special/interactivemaps/InteractiveMapPageObject.java
Dyktus: Yes it is. There is only one request with URL maps.wikia-services.com and it is ponto request

RodriGomez:
LGTM

PR #623, opened 2015-04-16
Mercury test fix

tomasznapieralski:
New social bar is on top of smart banner, i added waiter for social bar to be hidden.


RodriGomez:
LGTM

PR #629, opened 2015-04-22
QAART-574 Fix for RTE_017_MoreMainTools

rochan:
https://wikia-inc.atlassian.net/browse/QAART-574
One of the issues to cause a fail is that if the tool modal is still opened, the focus mode layer would block the interaction of the clicking on the submit button.
Adding wait for the modal and the focus mode to disappear before proceeding of next action.
This should alleviate issues for RTE_017, RTE_018, and RTE_019


RodriGomez:
LGTM

PR #632, opened 2015-04-28
removing lohgame

justnpT:
this will stabilize push to top tests


ludwikkazmierczak:
Can you elaborate more? What is the problem with searching lohgame?

PR #637, opened 2015-05-04
ADEN-1989 Prepare better selenium tests for Mercury

nandy-andy:
There are two issues currently with running tests for Mercury:
* `useskin=mercury` doesn't work in staging environments, so you can't run tests for this skin for instance on sandboxes,
* while developing we work on local instance of Mercury using [xipio](http://xip.io/) - we can't verify if our changes don't introduce regression because it's impossible to run selenium tests on our local instances of Mercury.

In the changes below I changed `UrlBuilder` in the way which allows us to solve those issues. After merging those changes we should be able to run selenium tests on a local Mercury instance by using `-Dbrowser=CHROMEMOBILEMERCURY` parameter along with `env` parameter set to a devbox name.

Example:
`mvn test -Dgroups=MercuryArticleTests -Dbrowser=CHROMEMOBILEMERCURY -Denv=dev-nandy`

//cc @drets @tomasznapieralski @ludwikkazmierczak

### DIFF
--- a/src/test/java/com/wikia/webdriver/common/core/urlbuilder/UrlBuilder.java
+++ b/src/test/java/com/wikia/webdriver/common/core/urlbuilder/UrlBuilder.java
@@ -10,6 +10,11 @@ import org.apache.commons.lang3.tuple.Pair;
  */
 public class UrlBuilder {
 
+  final String XIPIO_ADDRESS_FORMAT = ".%s:%d";
+
+  final String XIPIO_DEFAULT_DOMAIN = "127.0.0.1.xip.io";
+  final int XIPIO_DEFAULT_PORT = 8000;
+
   private String browser;
   private String env;
 
@@ -34,21 +39,28 @@ public class UrlBuilder {
   public String getUrlForPath(String wikiName, String wikiPath) {
     String url = getUrlForWiki(wikiName);
     String separator = wikiName.endsWith("wikia") || wikiName.equals("wowwiki") ? "" : "wiki/";
+
     if ("CHROMEMOBILE".equalsIgnoreCase(browser)) {
       return appendQueryStringToURL(url, "useskin=wikiamobile");
     }
+
     return url + separator + wikiPath;
   }
 
   public String getUrlForWiki(String wikiName) {
     String prefix = getUrlPrefix(wikiName);
     String suffix = getUrlSuffix(wikiName);
+
     if (customWikiNames.containsKey(wikiName)) {
-      prefix = suffix = "";
-      wikiName = env.contains("dev") ?
-                 (String) customWikiNames.get(wikiName).getRight() :
-                 (String) customWikiNames.get(wikiName).getLeft();
+      if (env.contains("dev")) {
+        prefix = "";
+        wikiName = (String) customWikiNames.get(wikiName).getRight();
+      } else {
+        prefix = suffix = "";
+        wikiName = (String) customWikiNames.get(wikiName).getLeft();
+      }
     }
+
     return composeUrl(prefix, wikiName, suffix);
   }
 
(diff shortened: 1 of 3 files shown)


src/test/java/com/wikia/webdriver/common/core/urlbuilder/UrlBuilder.java
```diff
@@ -10,6 +10,11 @@
  */
 public class UrlBuilder {
 
+  final String XIPIO_ADDRESS_FORMAT = ".%s:%d";
```
ludwikkazmierczak: A `private` and `static` would be nice

src/test/java/com/wikia/webdriver/common/core/urlbuilder/UrlBuilder.java
```diff
@@ -10,6 +10,11 @@
...
+
```
ludwikkazmierczak: Why we need to provide an environment to UrlBuilder()? Shouldn't we take this data from configuration? How about  actually using a `Builder` pattern. En example usage in my head looks like this:
```
UrlBuilder.getUrlForWiki("MyCoolWikia").forEnv("preview").forBrowser("CHROME").build();
```
and a default one should use a browser and environment provided from ConfigurationFactory.class, so nothing changes in a regular usage:
```
UrlBuilder.getUrlForWiki("MyCoolWikia").build();
```

src/test/java/com/wikia/webdriver/common/core/urlbuilder/UrlBuilder.java
```diff
@@ -64,22 +74,33 @@ private String getUrlPrefix(String wikiName) {
   }
 
   private String getUrlSuffix(String wikiName) {
+    if (env.contains("dev") && isMercuryBrowser()) {
+      return String.format(XIPIO_ADDRESS_FORMAT, XIPIO_DEFAULT_DOMAIN, XIPIO_DEFAULT_PORT);
```
drets: I would go with:
```java
return ".127.0.0.1.xip.io:8000";
``` 
and get rid of the 3 static vars. 
It would be more readable for me.

src/test/java/com/wikia/webdriver/testcases/adstests/TestUrlBuilder.java
```diff
@@ -123,6 +167,8 @@ public void testUrlBuilder() {
           .getUrlForPath((String) data[0], (String) data[1]));
       Assertion.assertEquals((String) data[5], new UrlBuilder("dev-dmytror")
           .getUrlForPath((String) data[0], (String) data[1]));
+        Assertion.assertEquals((String) data[6], new UrlBuilder("dev-nandy", "CHROMEMOBILEMERCURY")
```
drets: :+1:

drets:
The `TestUrlBuilder` extended and passes, so, LGTM!
Nice job @nandy-andy :wink:

src/test/java/com/wikia/webdriver/common/core/urlbuilder/UrlBuilder.java
nandy-andy: The readability reason depends on reader, I guess ;) Since IMHO the names of constants are self-explanatory. The idea was to make it possible in future to pass custom domain and port.

We're using this consts only once, so it might be interpreted as over-engineering :) Although, I found using constants a good practice in Java. I'll leave it for now since we want to merge it ASAP and the discussion about constants vs explicit strings can go on and on and on :japanese_ogre:

drets: The comment is a suggestion since I wrote to you LGTM before. And you are right from my point of view it's over-engineering :wink:

PR #645, opened 2015-05-11
ADEN-2001 Update to page where we have an ad slot

drets:
We use `popularSites` data provider in the different tests. One of them is a test that checks integration with Krux. The test contains verification that we added Krux data to our Ad slot. So, my PR is about changing one URL from the page where we don't have Ad slot (search page on mobile) to the page which contains an Ad slot (video page).


nandy-andy:
LGTM :+1:

PR #649, opened 2015-05-19
ADEN-2036 Fix waitForIframeLoaded()

nandy-andy:
The test was "fixed" by targeting house ads for the synthetic pages.

But since the safe iframes are being used much more often it seems removing the custom JavaScript and using Selenium driver features instead to check if an iframe is loaded might help. The changes below aren't connected only to the `TestLeaderboardAndPrefooterOnConsecutivePageViews()` test. I'm removing there custom JS and replacing it with the driver's features to switch to the iframe, wait for it to load and then switch back to the default content.

//cc @drets @ludwikkazmierczak

### DIFF
--- a/src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/adsbase/AdsBaseObject.java
+++ b/src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/adsbase/AdsBaseObject.java
@@ -1,7 +1,5 @@
 package com.wikia.webdriver.pageobjectsfactory.pageobject.adsbase;
 
-import com.google.common.base.Joiner;
-
 import com.wikia.webdriver.common.contentpatterns.AdsContent;
 import com.wikia.webdriver.common.core.Assertion;
 import com.wikia.webdriver.common.core.CommonExpectedConditions;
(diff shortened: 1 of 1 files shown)


src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/adsbase/AdsBaseObject.java
```diff
@@ -542,17 +541,11 @@ public void verifyNoLiftiumAdsInSlots(List<String> slots) {
...
-        "var iframe = arguments[0];" +
-        "if (iframe.contentWindow.document.readyState === 'complete'){ return callback(); } else {"
-        +
-        "iframe.contentWindow.addEventListener('load', function () {return callback(); }) " +
-        "}",
-        iframe
-    );
+    PageObjectLogging.log("waitForIframeLoaded", "Switching to adslot iframe", true);
+    driver.switchTo().frame(iframe);
+    waitPageLoaded();
```
drets: :+1: clever solution

drets:
LGTM

PiotrGackowski:
LGTM

PR #650, opened 2015-05-20
Improvements

tomasznapieralski:
Summary of changes:
- Add overloaded method ``public boolean areFilesTheSame(File file1, File file2, double accuracy){...}``
- change return statement to inline version in ``ImageComparison.java``
- change Mercury test to use new method

### DIFF
--- a/src/test/java/com/wikia/webdriver/testcases/mercurytests/LightboxTests.java
+++ b/src/test/java/com/wikia/webdriver/testcases/mercurytests/LightboxTests.java
@@ -101,8 +101,7 @@ public class LightboxTests extends NewTestTemplate {
   }
 
   // MT03
-  @Test(groups = {"MercuryLightboxTest_003", "MercuryLightboxTests",
-                  "Mercury"}, invocationCount = 1)
+  @Test(groups = {"MercuryLightboxTest_003", "MercuryLightboxTests", "Mercury"})
   public void MercuryLightboxTest_003_ZoomByGesture_ZoomByDoubleTap() {
     BasePageObject base = new BasePageObject(driver);
     base.openMercuryArticleByName(wikiURL, MercuryArticles.GALLERY);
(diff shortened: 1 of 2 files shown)


tomasznapieralski:
@ludwikkazmierczak @RodriGomez please take a look

src/test/java/com/wikia/webdriver/testcases/mercurytests/LightboxTests.java
```diff
@@ -99,7 +101,8 @@ public void MercuryLightboxTest_002_TapOnEdgesChangeImages_SwipeChangeImages() {
   }
 
   // MT03
-  @Test(groups = {"MercuryLightboxTest_003", "MercuryLightboxTests", "Mercury"})
+  @Test(groups = {"MercuryLightboxTest_003", "MercuryLightboxTests",
+                  "Mercury"}, invocationCount = 1)
```
RodriGomez: why is invocationCount = 1?

tomasznapieralski: Good catch i forgot to remove that after debugging

src/test/java/com/wikia/webdriver/testcases/mercurytests/LightboxTests.java
```diff
@@ -112,11 +114,13 @@ public void MercuryLightboxTest_003_ZoomByGesture_ZoomByDoubleTap() {
     touchAction.zoomInOutPointXY(50, 50, 50, 100, PerformTouchAction.ZOOM_WAY_IN, 3000);
     File afterZooming = new Shooter().capturePage(driver);
```
drets: What do you think about converting `File` to `BufferedImage`, since it is a screenshot from viewPort and to use `ImageComparison::areImagesDifferent` method. 
From my point of view, it's more transparent to use comparison of pixels in images than bytes in files.

tomasznapieralski: I will try

tomasznapieralski: For that test we don't need to get such detailed comparison results.
For that tests steps are:
- capture image 1
- zoom in
- capture image 2
- compare 1 and 2, if they are different mark zoom in as pass
- zoom out
- capture image 3
- compare 1 and 3, if they are the same mark zoom out as pass

Byte difference for that test is nearly doubled:
- zoom out about 250kb
- zoom in about 500kb

So checking byte lenght is fine for that

PR #652, opened 2015-05-25
ADEN-2040 Fix user logging

drets:


src/test/java/com/wikia/webdriver/testcases/adstests/TestNoAdsLoggedInUsers.java
```diff
@@ -35,11 +35,9 @@ public TestNoAdsLoggedInUsers(String wikiName, String path) {
   }
 
   private void loginSteps() {
```
nandy-andy: Why is it call this way? :japanese_ogre: I can't see any "steps" there ;) Wouldn't it be easier to call it just `login()`? :)

nandy-andy:
LGTM

src/test/java/com/wikia/webdriver/testcases/adstests/TestNoAdsLoggedInUsers.java
drets: It came from another person, but I will change it for you and for clearness. :dancers:

PR #656, opened 2015-05-26
Small Mercury fix

tomasznapieralski:


src/test/java/com/wikia/webdriver/testcases/mercurytests/SEOTests.java
```diff
@@ -24,12 +24,14 @@ public void prepareTest() {
     wikiURL = urlBuilder.getUrlForWiki("muppet");
   }
 
+  private static final String MUPPET_MAIN_PAGE = "Muppet_Wiki";
```
RodriGomez: Why not add this to MercuryArticles.java?

tomasznapieralski: Because MercuryArticles.java has only articles within MercuryAutomationTesting Wiki

src/test/java/com/wikia/webdriver/testcases/mercurytests/SmartBannerTests.java
```diff
@@ -22,14 +22,19 @@
...
+   */
+  private static final String[][] WIKIS = {
+      {"destiny", "Destiny_Wiki", "#94d11f"},
+      {"cocktails", "Cocktails_Wiki", "#8ca038"},
+      {"thehungergames", "The_Hunger_Games_Wiki", "#ff7f26"},
+      {"dc", "Main_Page", "#ff5400"},
+      {"tardis", "Doctor_Who_Wiki", "#00b7e0"},
+      {"starwars", "Main_Page", "#09d3bf"},
+      {"lego", "LEGO_Wiki", "#ffd000"}
+  };
```
RodriGomez: Did you think about using DataProvider instead?

tomasznapieralski: Sometimes i will need two or more packages in the same test that is why i cannot use DataProvider. From the other hand in MercuryArticles.java, i'm calling only few of all articles in the same test using static fields.

src/test/java/com/wikia/webdriver/testcases/mercurytests/SmartBannerTests.java
```diff
@@ -22,14 +22,19 @@
...
-  private static final String[]
-      DIFFERENT_HUBS_COLORS =
-      {"#94d11f", "#8ca038", "#ff7f26", "#ff5400", "#00b7e0", "#09d3bf",
-       "#ffd000"};
+
+  /**
+   * Wiki name, Main page, HUB color
+   */
+  private static final String[][] WIKIS = {
+      {"destiny", "Destiny_Wiki", "#94d11f"},
```
drets: What do you think if instead of using hex color values as string make an `enum` which will have color names and hex color values?
It will be more readable and reusable :wink:

tomasznapieralski:
Added enum Colors

PR #657, opened 2015-05-27
Fix for Facebook automation defects SOC-823

Archbold:
Fix for Facebook automation defects SOC-823
NB this fixes only fixes SOC-823 not SOC-774 my mistake, so ignore branch name and please CR and merge.

### DIFF
--- a/src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/facebook/FacebookSettingsPageObject.java
+++ b/src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/facebook/FacebookSettingsPageObject.java
@@ -32,7 +32,7 @@ public class FacebookSettingsPageObject extends WikiBasePageObject {
   private WebElement settingsContent;
   @FindBy(css = "#userNavigationLabel")
   private WebElement fbDropDown;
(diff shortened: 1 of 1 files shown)


Archbold:
NB this fixes only fixes SOC-823 not SOC-774 my mistake, but please CR and merge.

src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/facebook/FacebookSettingsPageObject.java
```diff
@@ -32,7 +32,7 @@
   private WebElement settingsContent;
   @FindBy(css = "#userNavigationLabel")
   private WebElement fbDropDown;
-  @FindBy(css = ".uiLinkButtonInput")
+  @FindBy(xpath = "//*[contains(@class, 'uiContextualLayer')]//span[text()='Log out']")
```
ukaka: ```xpath = "//*[contains(@class, 'uiContextualLayerBelowRight')]//span[text()='Log out']"``` would be even better selector in this case, because there is only one element with 'uiContextualLayerBelowRight' class and three elements with 'uiContextualLayer' class. So in case there will be other element with text "Log out", it will pass.

ukaka:
LGTM

PR #661, opened 2015-06-01
ADEN-2061 Floor Adhesion tests

nandy-andy:
Test for Oasis and Mercury to check basic functionality of Floor Adhesion ad unit logic.

More about Floor Adhesion: https://one.wikia-inc.com/wiki/Ad_Engineering/Ads/Floor_adhesion


src/test/java/com/wikia/webdriver/testcases/adstests/TestFloorAdhesion.java
```diff
@@ -0,0 +1,90 @@
...
+    private void verifyFloorAdhesionIsVisible(AdsBaseObject wikiPage) {
+        wikiPage.waitForElementByCss(FLOOR_ADHESION_SELECTOR);
+        PageObjectLogging.log(
+                "testFloorAdhesionOnOasis",
+                "Floor Adhesion unit visible",
+                true
+        );
+    }
+
+    private void verifyClickingUnitOpensLightbox(
```
ludwikkazmierczak: IMO you should move this selectors and and "verify" methods to some other class. And a better practice from my point of view is - instead of "verifyMyFancyThingIsThere" you should go with:

```
PageObjectLogging.log(
        "Check visiblity", 
        "Lightbox should be opened after clicking",
        page.clikUnit().isLightBoxOpened()
)
```

src/test/java/com/wikia/webdriver/testcases/adstests/TestFloorAdhesion.java
```diff
@@ -0,0 +1,90 @@
...
+    private final String
+            FLOOR_ADHESION_CLOSE_SELECTOR = "#ext-wikia-adEngine-template-footer .close";
+    private final String WIKIA_BAR_SELECTOR = "#WikiaBar";
+
+    @Test(
+            dataProvider = "floorAdhesionOnOasis",
+            dataProviderClass = AdsDataProvider.class,
+            groups = {"TestFloorAdhesionOasis"}
+    )
+    public void testFloorAdhesionOnOasis(
```
ludwikkazmierczak: Is @DataProvider really needed? Any extending comming in the nearest future?

src/test/java/com/wikia/webdriver/testcases/adstests/TestFloorAdhesion.java
```diff
@@ -0,0 +1,90 @@
...
+import com.wikia.webdriver.common.templates.TemplateDontLogout;
+import com.wikia.webdriver.pageobjectsfactory.pageobject.adsbase.AdsBaseObject;
+import com.wikia.webdriver.pageobjectsfactory.pageobject.adsbase.mobile.MobileAdsBaseObject;
+
+import org.openqa.selenium.By;
+import org.openqa.selenium.WebElement;
+import org.testng.annotations.Test;
+
+public class TestFloorAdhesion extends TemplateDontLogout {
+    private final String FLOOR_ADHESION_SELECTOR = "#ext-wikia-adEngine-template-footer";
```
ludwikkazmierczak: I would rename this to `FLOOR_ADHESION_CSS` - it's more meaning full then "selector" (can be XPath selector). but this is just IMO :)

src/test/java/com/wikia/webdriver/testcases/adstests/TestFloorAdhesion.java
```diff
@@ -0,0 +1,90 @@
...
+    private final String FLOOR_ADHESION_AD_SELECTOR = "#ext-wikia-adEngine-template-footer .ad";
+    private final String
+            FLOOR_ADHESION_CLOSE_SELECTOR = "#ext-wikia-adEngine-template-footer .close";
+    private final String WIKIA_BAR_SELECTOR = "#WikiaBar";
```
ludwikkazmierczak: ofc, comment above is applicable for all these fields

src/test/java/com/wikia/webdriver/testcases/adstests/TestFloorAdhesion.java
gabrys: Selector is "just" either XPath or CSS selector. CSS would mean the stylesheet or the CSS URL to me (before I think it can be also a CSS selector)

drets:
I would automate the following test cases:
  1. Check a Floor Adhesion in a page:
    * Check a Floor Adhesion selector
  2. Check a Floor Adhesion an ad image:
    * Get Floor Adhesion an ad web element image and compare with the template image
  3. Check close button of Floor Adhesion:
    * Click a Floor Adhesion close button
    * Check a Floor Adhesion is not visible
  4. Check a Floor Adhesion ad is in the bottom of a page:
    * Check a Floor Adhesion ad is in the bottom of a page
  5. Check a Floor Adhesion Lightbox:
    * Click a Floor Adhesion ad
    * Check a Floor Adhesion Lightbox selector
  6. Check a Floor Adhesion Lightbox an ad image:
    * Click a Floor Adhesion ad
    * Get a Floor Adhesion Lightbox ad web element image and compare with the template image
  7. Check a Floor Adhesion Lightbox close button:
    * Click a Floor Adhesion ad
    * Click a Floor Adhesion Lightbox close button
    * Check a Floor Adhesion Lightbox is not visible
    * Check a Floor Adhesion is visible
  8. Check a Floor Adhesion Lightbox ad is in the centre of a page:
    * Click a Floor Adhesion ad
    * Check a Floor Adhesion Lightbox ad is in the centre of a page

src/test/java/com/wikia/webdriver/testcases/adstests/TestFloorAdhesion.java
gabrys: I think we generally use CSS selectors in selenium tests, so I'm fine with naming the fields _SELECTOR.

src/test/java/com/wikia/webdriver/testcases/adstests/TestFloorAdhesion.java
gabrys: AFAIU it's only used as a way to parametrize the test for mobile/desktop. We could have separate methods for both skins launching the original test, but this is what data provider is doing, so we re-used the mechanism ;-).

src/test/java/com/wikia/webdriver/testcases/adstests/TestFloorAdhesion.java
drets: +1 for _SELECTOR

src/test/java/com/wikia/webdriver/testcases/adstests/TestFloorAdhesion.java
ludwikkazmierczak: so maybe `..._CSS_SELECTOR` :P When writing a test it is easier to see your own mistake (or during code reviewing) when using:
```
webdriver.findElement(By.cssSelector(MY_UBER_CSS));
webdriver.findElement(By.xpath(MY_UBER_XPATH));
```
instead of:
```
webdriver.findElement(By.xpath(MY_UBER_SELECTOR));
webdriver.findElement(By.cssSelector(MY_UBER_SELECTOR));

```

gabrys: Yes, _CSS_SELECTOR makes sense as well :-)

PiotrGackowski:
Please add slot to "setSlotsSelectors" method in "AdsContent" class. So we will be checking this slot during "no ads" tests.

src/test/java/com/wikia/webdriver/testcases/adstests/TestFloorAdhesion.java
nandy-andy: Thanks for the hint with `PageObjectLogging.log()` I'll implement it right away! :) 

I'm not sure about having the selectors and those methods in other class. @PiotrGackowski suggested `AdsBaseObject` or `MobileAdsBaseObject`. I'm not a big fan of this solution unless other test classes use them - right now there are no other test classes which need either the selectors nor the methods. What are the reasons to move the stuff somewhere else?

src/test/java/com/wikia/webdriver/testcases/adstests/TestFloorAdhesion.java
nandy-andy: Yep, that's correct. The `TestFloorAdhesion` test isn't skin-dependent. With the right data provided it works for both Oasis and Mercury. Not using data provider here will result in having two separate tests for both skins with a lot of code duplication or having one test with a lot of if-skin logic which is :no_good:

src/test/java/com/wikia/webdriver/testcases/adstests/TestFloorAdhesion.java
nandy-andy: Sure, I can go with `_CSS_SELECTOR`, I just saw `_SELECTOR` in other tests both for XPath and CSS, so I wanted to be consistent ;)

src/test/java/com/wikia/webdriver/testcases/adstests/TestFloorAdhesion.java
nandy-andy: Yep, this is what I wrote above :smile_cat:

nandy-andy:
@PiotrGackowski as discussed offline: there are no new slot introduced with this ad unit. More: check the documentation (and "code review" it also please :smile_cat:)

However, `INVISIBLE_HIGH_IMPACT_SLOT` wasn't added to the method and it's being used by the unit on mobile. But I can't find `setSlotSelectors` method in `AdsContent` class. Also, I'm not sure if it's OK to add the slot there now when it's available only on some communities.

src/test/java/com/wikia/webdriver/testcases/adstests/TestFloorAdhesion.java
ludwikkazmierczak: Test is for testing - PageObject is for handling page interaction - for keeping "The Single Responsibility Principle". After talking to Nandy I recomended to return objects based on environment, this way we can have a single test for both cases :)

drets:
It looks fine to me :smile_cat:

PR #662, opened 2015-06-02
ADEN-1684 Add ATF tests

drets:


src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/adsbase/AdsBaseObject.java
```diff
@@ -624,4 +627,31 @@ public void checkSkin(String adSkinLeftPath, String adSkinRightPath) {
...
+      restoreDeaultImplicitWait();
+    }
+  }
+
+  private void triggerFloatingMedrec(WebElement slot) {
+    executeScript("window.scroll(0, 5000); setTimeout(function () {window.scroll(0, 5001) }, 500)");
+    // try to wait a floating medrec 3 seconds
+    try {
+      new WebDriverWait(driver, 3).until(CommonExpectedConditions.elementVisible(slot));
+    } catch (org.openqa.selenium.TimeoutException ignore) {
```
ludwikkazmierczak: en empty catch seems strange here - maybe return false when exception is present or log a message? I think this part is just for triggering, you are not interested if you hide or show floating medrec, am I right? Or floating medrec is expected to be visible after triggering?

src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/adsbase/AdsBaseObject.java
```diff
@@ -624,4 +627,31 @@ public void checkSkin(String adSkinLeftPath, String adSkinRightPath) {
     PageObjectLogging.log("SKIN", "SKIN presents on the page", true);
   }
 
+  /**
+   * Check if AdEngine loaded the ad web elements inside slot.
+   */
+  public boolean checkIfSlotOnPageLoaded(String slotName) {
```
ludwikkazmierczak: `isSlotOnPageLoaded(...` would be nice

src/test/java/com/wikia/webdriver/testcases/adstests/TestATF.java
```diff
@@ -0,0 +1,79 @@
...
+    Assertion.assertTrue(adsBaseObject.checkIfSlotOnPageLoaded(AdsContent.MEDREC));
+    Assertion.assertTrue(adsBaseObject.checkIfSlotOnPageLoaded(AdsContent.TOP_LB));
+    Assertion.assertTrue(adsBaseObject.checkIfSlotOnPageLoaded(AdsContent.INVISIBLE_SKIN));
+
+    if (isWgVarOn) {
+      Assertion.assertFalse(adsBaseObject.checkIfSlotOnPageLoaded(AdsContent.PREFOOTER_LEFT));
+      Assertion.assertFalse(adsBaseObject.checkIfSlotOnPageLoaded(AdsContent.PREFOOTER_RIGHT));
+      Assertion.assertFalse(adsBaseObject.checkIfSlotOnPageLoaded(AdsContent.LEFT_SKYSCRAPPER_3));
+      Assertion.assertFalse(adsBaseObject.checkIfSlotOnPageLoaded(AdsContent.FLOATING_MEDREC));
+      Thread.sleep(1000 * delaySec);
```
ludwikkazmierczak: You are waiting for this vars to change after some delay, is this delay part of functionality?

src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/adsbase/AdsBaseObject.java
drets: The method tries to trigger a floating Medrec and waits 3 seconds till slot will be loaded.
I used `triggerFloatingMedrec` inside `isSlotOnPageLoaded` which can return false if a slot is not loaded. So, it could be the case when we try to trigger medrec, but configuration is not allow to do it. Should I rename to `tryTriggerFloatingMedrec`?

src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/adsbase/AdsBaseObject.java
drets: Done.

src/test/java/com/wikia/webdriver/testcases/adstests/TestATF.java
drets: The scenario is following:
(if wgVar is set to true for wiki)
* check that slots are not shown
* wait 20 sec
* check that slots are appeared

I configured the ads in this page in such a way that after 20 sec it should appear.
I don't see how can I use explicit wait here. I am fine with `Thread.sleep` because delay is part of functionality.

src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/adsbase/AdsBaseObject.java
ludwikkazmierczak: you should log each situation with a separate message, or throw a meaningful exception - `FloatingMedrecNotAllowed`, cause if you want to trigger this element, and it's not available - it's an unexpected situation ASAIU

drets: Changed method name.

nandy-andy:
It looks fine to me :smile_cat:

PR #663, opened 2015-06-03
CONCF-620 remove tests related with share toolbar

bognix:
ping @justnpT @ludwikkazmierczak @Warkot


bognix:
Tests don't have to be removed from Jenkins wikia-test-matrix because it is inside "Toolbar" group so it is going to be removed automatically.

justnpT:
LGTM

Warkot:
I'm not a Selenium master, but it LGTM.

PR #665, opened 2015-06-05
ADEN-2044 Amazon on Mercury tests

nandy-andy:
Since we want to rollout AmazonMatch to Mercury we want an automation test for it. I have added test in https://github.com/Wikia/selenium-tests/commit/be86e21e58712ae5f0a8bad523f8b9fbb52dc80a. The second commit is reformatting code.


drets:
LGTM

PR #672, opened 2015-06-09
Create more tests for mercury

tomasznapieralski:
This is beginning, i will update status using labels.
Please don't merge it i will do this when time comes.


tomasznapieralski:
@RodriGomez please CR

tomasznapieralski:
If you will have any objections about that i will create PR with fix

PR #677, opened 2015-06-11
ADEN Update provider chain tests

drets:


nandy-andy:
No GeoEdge which is such a pain recently... Yay :dancer:

LGTM :+1:

PR #678, opened 2015-06-11
SOC-843 add logic to reset forgot password time

bekcunning:
added logic that was initially left out to reset the forgot password timing so that a user can request new password multiple times in a 24 hour period.


ludwikkazmierczak:
LGTM

PR #679, opened 2015-06-11
SOC-774 removed BeforeGroup and AfterGroup now that preview FB app ha…

bekcunning:
a new FB app is being used for preview so that the callback is working properly when a user removes the app from facebook settings.  This means we no longer need Before and AfterMethods to clean up the test accounts.  I've updated this test case to reflect that it will now work properly and no longer needs to depend on additional cleanup steps.


ludwikkazmierczak:
LGTM

PR #688, opened 2015-06-16
ADEN Reduce krux urls

drets:


nandy-andy:
:red_circle: LGTM :+1:

PR #695, opened 2015-06-22
updated logout selector to reflect current Facebook selectors

Archbold:


src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/facebook/FacebookSettingsPageObject.java
```diff
@@ -32,7 +32,7 @@
   private WebElement settingsContent;
   @FindBy(css = "#userNavigationLabel")
   private WebElement fbDropDown;
-  @FindBy(css = "input.uiLinkButtonInput")
+  @FindBy(xpath = "//*[contains(@class, 'uiContextualLayerBelowRight')]//span[text()='Log out']")
```
ludwikkazmierczak: `//span[text()='Log out']` wouldn't this be enough?

PR #702, opened 2015-06-24
ADEN-2156 ATF fix

drets:


nandy-andy:
We changed that because with ADEN-2066 we've introduced one more div wrapper.

LGTM :+1:

PR #704, opened 2015-06-25
ADEN-2173 Check floor adhesion against line item id and creative id

gabrys:


src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/adsbase/helpers/AdsFloorAdhesionSkinContext.java
```diff
@@ -0,0 +1,57 @@
+package com.wikia.webdriver.pageobjectsfactory.pageobject.adsbase.helpers;
+
+public class AdsFloorAdhesionSkinContext {
+    private final String OASIS_CREATIVE_ID = "66223683972";
```
drets: I would change it from `private final` to `private static` however it came from old code.

drets:
LGTD

PR #705, opened 2015-06-25
Two generic fixes from ad engineering

gabrys:
This pull requests adds two fixes to this repo.

Firstly the HTML dumps generated by PageObjectLogging include the scripts, which causes them to be run again (even though the effects of them are already applied to the HTML dump). In my change I wrap them in invisible textareas, so they are not run again, but if you need to see what they were, you can find them using Chrome or Firefox inspector.

The second change is how Assertion.assertEquals is called. It used to be:

    Assertion.assertEquals(String pattern, String current);

I changed that to:

    Assertion.assertEquals(String current, String pattern);

This matches org.testng.Assert:

    Assert.assertEquals(String actual, String expected);

Also this fixes the issue with reporting the pattern and current value in the HTML log.

The remaining Assertion.assert* methods were updated to match this convention as well (see a5a69c10fce)

### DIFF
--- a/src/test/java/com/wikia/webdriver/common/core/Assertion.java
+++ b/src/test/java/com/wikia/webdriver/common/core/Assertion.java
@@ -34,12 +34,12 @@ public class Assertion extends Assert {
     return assertion;
   }
 
-  public static void assertEquals(String pattern, String current) {
+  public static void assertEquals(String current, String pattern) {
     String patternEncoded = encodeSpecialChars(pattern);
     String currentEncoded = encodeSpecialChars(current);
     boolean assertion = true;
     try {
-      Assert.assertEquals(pattern, current);
+      Assert.assertEquals(current, pattern);
     } catch (AssertionError err) {
       addVerificationFailure(err);
       assertion = false;
--- a/src/test/java/com/wikia/webdriver/common/core/Assertion.java
+++ b/src/test/java/com/wikia/webdriver/common/core/Assertion.java
@@ -13,7 +13,7 @@ import java.util.Map;
 
 public class Assertion extends Assert {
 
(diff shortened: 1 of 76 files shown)


src/test/java/com/wikia/webdriver/common/logging/PageObjectLogging.java
```diff
@@ -52,11 +52,17 @@
   private By lastFindBy;
   private WebDriver driver;
 
+  private static String getPageSource(WebDriver driver) {
+    return getPageSource(driver)
+            .replaceAll("<script", "<textarea style=\"display: none\"><script")
+            .replaceAll("</script", "</script></textarea");
```
tomasznapieralski: Was missing `>` intentional?

gabrys: Yes

drets:
LGTM

tomasznapieralski:
LGTM

src/test/java/com/wikia/webdriver/common/core/Assertion.java
```diff
@@ -34,12 +34,12 @@ public static boolean assertStringContains(String pattern, String current) {
...
   }
 
-  public static void assertEquals(String pattern, String current) {
+  public static void assertEquals(String current, String pattern) {
     String patternEncoded = encodeSpecialChars(pattern);
     String currentEncoded = encodeSpecialChars(current);
     boolean assertion = true;
     try {
-      Assert.assertEquals(pattern, current);
+      Assert.assertEquals(current, pattern);
```
drets: Good catch :+1:

PR #715, opened 2015-06-30
ADEN-1932 Test for there's no skin on small screen

gabrys:


drets:
LGTM

justnpT:
I've just noticed that the AdsDataProvider has an empty constructor

    private AdsDataProvider() {

    }

It is also a very large file already
Has the pottential for small refactoring

LGTM

PR #718, opened 2015-07-01
Concf 767

justnpT:
https://wikia-inc.atlassian.net/browse/CONCF-767

* Create test which tries to save CC without an image
* Create test which saves CC with an image successfully

The test assumes that error message on image will be extended with error class. The class is used for verification of the error message.

In class SpecialCuratedContentPageObject I focused methods and webelement on the last element, on the page because most automated test scenarios will operate on the last element. The solution is simple when compared to alternative approach which uses list of all elements and operates on them in different loops. If any engineer would ever like to operate on elements other than the last one, he can create such methods. However i believe all atomated test scenarios in CuratedContent can be performed by operating only on the last element.

Please don't merge according to the Content Team board, which means the test must be verified after the code review, when the functionality is introduced.

### DIFF
--- a/src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/special/SpecialCuratedContentPageObject.java
+++ b/src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/special/SpecialCuratedContentPageObject.java
@@ -9,8 +9,7 @@ import org.openqa.selenium.support.FindBy;
 import org.openqa.selenium.support.PageFactory;
 
 /**
- * @author Garth Webb
- * @author Saipetch Kongkatong
+ * @ownership Content
  */
 public class SpecialCuratedContentPageObject extends SpecialPageObject {
 
@@ -19,7 +18,7 @@ public class SpecialCuratedContentPageObject extends SpecialPageObject {
   @FindBy(css = "#save")
   private WebElement saveButton;
   @FindBy(css = "#save.ok")
-  private WebElement successMark;
+  private WebElement successfulSaveMark;
   @FindBy(css = "li.item:last-of-type .item-input")
   private WebElement lastElementNameInput;
   @FindBy(css = "li.item:last-of-type .name")
@@ -54,7 +53,7 @@ public class SpecialCuratedContentPageObject extends SpecialPageObject {
   }
 
   public void verifySuccesfulSave() {
-    waitForElementByElement(successMark);
+    waitForElementByElement(successfulSaveMark);
     PageObjectLogging.log("verifySuccesfulSave", "Curated Content saved with success", true);
   }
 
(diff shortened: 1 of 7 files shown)


src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/special/SpecialCuratedContentPageObject.java
```diff
@@ -0,0 +1,85 @@
...
+  @FindBy(css = "li.item:last-of-type .image")
+  private WebElement lastElementImage;
+  @FindBy(css = "li.item:last-of-type .image[style]")
+  private WebElement lastElementImageFilled;
+  @FindBy(css = "li.item:last-of-type .image.error")
+  private WebElement lastElementImageWithError;
+
+  public SpecialCuratedContentPageObject(WebDriver driver) {
+    super(driver);
+    PageFactory.initElements(driver, this);
```
tomasznapieralski: Your parent constructor is doing that for you `WikiBasePageObject::WikiBasePageObject()`

src/test/java/com/wikia/webdriver/testcases/specialpagestests/CuratedContentTests.java
```diff
@@ -0,0 +1,56 @@
...
+
+import org.testng.annotations.Test;
+
+/**
+ * @ownership Content
+ */
+public class CuratedContentTests extends NewTestTemplate {
+
+  String category = URLsContent.CATEGORY_HELP;
+  String label = PageContent.LOREM_IPSUM_SHORT;
```
tomasznapieralski: Please use private static final for both

tomasznapieralski:
Use capital letter for test methods like `public void CuratedContent002_saveWithImage()`

tomasznapieralski:
When you are waiting for element which intentionally won't be visible use:
`waitForElementVisibleByElement(WebElement element, int timeOutInSec, int checkOutInMilliSec)`
because by default you will wait 30 sec for element and in that method you can reduce this time to i.e. 5 sec. example: `waitForElementVisibleByElement(WebElement element, 5, 1000)`

tomasznapieralski:
please follow pattern that verify method should be splited into checking functionality and reporting it's status. For example:
```
In page/component object:
  public boolean isSaveSuccessful() {
    try {
      waitForElementVisibleByElement(successfulSaveMark, 5, 1000);
    } catch (TimeoutException e) {
      return false;
    }
    return true;
  }

In test class:
  public void curatedContent002_saveWithImage() {
    ...
    PageObjectLogging.log("verifySuccesfulSave", "Curated Content saved with success", cc.isSaveSuccessful());
    ...
  }
```

tomasznapieralski:
CR first round done

tomasznapieralski:
LGTM

PR #719, opened 2015-07-02
Configuration refactor

ludwikkazmierczak:
Replacing an over-complicated configuration management with a single point of configuration. Removed maven profiles handling manual configuration. Configuration property handling looks like this now: 
1. If there is a System Property provided (like in console run: "-DsomeProp=someValue")
2. If no System Property is found - property is provided from configuration files (config_sample.yml if no config.yml provided)
NOTE: system Property overrides ANY property

This is an initial change - next step will be 
1. unify configuration spread over  xml, and yml files to a single format 
2. move config to resources and provide an easy-to-use mechanism for overriding default properties for your local runs (kind of what we do with config_sample.yml and config.yml)
3. refactor Credentials to be easier to use

 AbstractConfiguration.java,  ConfigurationFactory.java,  ManualConfiguration.java and POMConfiguration.java classes are replaced with single Configuration.java class

How it affects the testing:
Instead of:
`ConfigurationFactory.getConfig().getBrowser();`
 you are now using just: 
`Configuration.getBrowser();`
same happens with ALL other properties, you can now access them using Credentials class.

NOTE: No properties are now held by pom.xml file
NOTE: No difference in configuration cause of running with IDE not console! Finally!


tomasznapieralski:
Nice work!

src/test/java/com/wikia/webdriver/common/core/configuration/Configuration.java
```diff
@@ -0,0 +1,107 @@
...
+  public static String getDisableFlash() {
+    return String.valueOf(getProp("disableFlash"));
+  }
+
+  public static Credentials getCredentials() {
+    return new Credentials(new File(getCredentialsFilePath()));
+  }
+
+  public static String getEnvType() {
+    if (getEnv().contains("prod")) {
```
drets: Potential `NullPointerException`

src/test/java/com/wikia/webdriver/common/core/configuration/Configuration.java
```diff
@@ -0,0 +1,107 @@
...
+    if (config == null) {
+      Yaml yaml = new Yaml();
+      InputStream input = null;
+      try {
+        input = new FileInputStream(new File("config.yml"));
+      } catch (FileNotFoundException ex) {
+        try {
+          input = new FileInputStream(new File("config_sample.yml"));
+        } catch (FileNotFoundException ex2) {
+          System.out.println("CAN'T LOCATE CONFIG FILE");
```
drets: throw exception?

src/test/java/com/wikia/webdriver/common/core/configuration/Configuration.java
```diff
@@ -0,0 +1,107 @@
...
+  public static String getAppiumIp() {
+    return getProp("appiumIp");
+  }
+
+  public static String getDeviceName() {
+    return getProp("deviceName");
+  }
+
+  public static String getDisableFlash() {
+    return String.valueOf(getProp("disableFlash"));
```
drets: Why do we need `String.valueOf`?

drets:
:+1:

config_sample.yml
```diff
@@ -2,11 +2,11 @@
 #Copy its content to config.yml file
 #It is used only when you run tests using IDE
 
-browser: FF #acceptable values: FF, CHROME, GHOST, ANDROID
+browser: CHROME #acceptable values: FF, CHROME, GHOST, ANDROID
```
nandy-andy: What about those fancy `CHROMEMOBILE` and `CHROMEMOBILEMERCURY` mentioned in the README? Is it something different? If not, I'd like them to be mentioned here in the comment as well ;)

ludwikkazmierczak: Done

src/test/java/com/wikia/webdriver/common/core/configuration/Configuration.java
ludwikkazmierczak: Not changing a logic right now, just a config point

config_sample.yml
nandy-andy: Thanks! :smile_cat:

PR #720, opened 2015-07-02
SOC-996 adding testcases for error cases in mobile registration

bekcunning:
Covered
-username in use
-password matches username
-email already taken
-too young user (birthdate error)


ludwikkazmierczak:
LGTM

PR #727, opened 2015-07-06
ADEN-2201 Small change in `TestLeaderboardAndPrefooterOnConsecutivePageViews`

nandy-andy:
While working on Amazon tests in Mercury we found out waiting for preloader in Mercury isn't good enough. We used this method (`mercuryWaitForPreloaderToHide()`) also in `TestLeaderboardAndPrefooterOnConsecutivePageViews` which started failing today on preview. We decided to get rid of suboptimal method and used `waitTitleChangesTo()` instead.


PiotrGackowski:
LGTM

PR #728, opened 2015-07-06
wait for element clickability before click

justnpT:
Lest be careful here and wait for the element clickability after we scroll to it.
The scroll may be very fast and timings are unknown so its more safe to wait for the element to be clickable. 

I suspect this wait to be the fix of a failure that I noticed today.
It's very rare failure but this fix may prevent it.

### DIFF
--- a/src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/BasePageObject.java
+++ b/src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/BasePageObject.java
@@ -221,7 +221,7 @@ public class BasePageObject {
 
   protected void scrollAndClick(WebElement element) {
     scrollToElement(element);
-    waitForElementClickableByElement(element);
+    waitForElementClickableByElement(element, 5);
     element.click();
   }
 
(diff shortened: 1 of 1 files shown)


src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/BasePageObject.java
```diff
@@ -221,6 +221,7 @@ protected String getPseudoElementValue(WebElement element, String pseudoElement,
 
   protected void scrollAndClick(WebElement element) {
     scrollToElement(element);
+    waitForElementClickableByElement(element);
```
ludwikkazmierczak: This one is using a default timeout, (thath is 30 sec) - maybe 5sec is enough?

justnpT: I'm wondering where and how the timetout change should be added then. 
Maybe we should  create method "changeWaitTimeout(int timeout)" that performs that and do this:

changeWaitTimeout(5);
waitForElementClickableByElement(element);
changeWaitTimeout(30);

justnpT: @ludwikkazmierczak what do you think

drets: I would go with:
```java
waitForElementClickableByElement(element, 5);
```
rather then:
```java
changeWaitTimeout(5);
waitForElementClickableByElement(element);
changeWaitTimeout(30);
```

justnpT:
@ludwikkazmierczak @drets how about this ? I introduced Dimo idea

drets:
@justnpT to be honest I don't see any problems for using first version of your code with default 30 sec timeout.
I like your idea for tracking the longest successful wait :+1: 
If you want to go using different timeouts I would write:
```java
  public void waitForElementClickableByElement(WebElement element, int newTimeOut) {
    new WebDriverWait(driver, newTimeOut).until(
        CommonExpectedConditions.elementToBeClickable(element));
  }
```
It's not obligatory to assign new `WebDriverWait` object to `wait` variable and do tricky things with implicit wait since for new `WebDriverWait` object implicit wait is 0.

drets:
LGTM

PR #754, opened 2015-07-17
Improved Forgotten password tests to delete emails before requesting password

Archbold:
Improved Forgotten password tests to delete emails before requesting forgotten password rather than afterwards ;)


RodriGomez:
LGTM

PR #756, opened 2015-07-17
CONCF-800 | add test CCT09

tomasznapieralski:
Algorithm:
* open http://mercurycc.wikia.com/main/category/Templates
* check that you are on category page
* open http://mercurycc.wikia.com/main/section/Categories
* check that you are on section page
* open http://mercurycc.wikia.com/main/section/qwerty
* check that you were redirected to /
* open http://mercurycc.wikia.com/main/category/qwerty
* check that you were redirected to /

PR #757, opened 2015-07-17
CONCF-802 | add test CCT08 and CCT10

tomasznapieralski:
Algorithm for CCT08:
* open http://mercurycc.wikia.com/main/category/Templates
* check that exactly 24 items are visible
* check that load more button is visible
* tap on load more button
* check that exactly 28 items are visible
* iterate through items and check that they are unique
* check that load more button is hidden

Algorithm for CCT10:
* open http://mercurycc.wikia.com/wiki/Category:Articles
* check that there is less than 24 items
* check that load more button is hidden


tomasznapieralski:
don't merge before https://github.com/Wikia/selenium-tests/pull/756

Warkot:
Wouldn't it be more errorproof if you create special categories with exactly expected items count like `24 items category`, `28 items category`? If anyone adds a template or an article on this wiki, tests will fail.

tomasznapieralski:
That is why testing environment shouldn't be changed, to avoid that kind of problems

Warkot:
I agree it shouldn't be changed, but I still disagree with this approach — anyone could do it by mistake, by adding new test cases or doing something else. It's harder to change a dedicated category, which is maintained manually than auto-created category where items are auto-added. Test cases should be as separated as possible and their purpose explicit.

tomasznapieralski:
fixed, thanks @justnpT for creating categories

Warkot:
LGTM

PR #769, opened 2015-07-22
ADEN-2248 Resize the browser window to 1900x900

gabrys:
Resize the browser to 1900x900 so we can run the tests on the regular
machines and not necessarily the high resolution ones.


drets:
LGTM

PR #853, opened 2015-09-11
Different email addresses were being used within the method. Changed …

Archbold:
Two different email addresses were being used within the method. Effect was that incorrect mail box was being looked in for the the forgotten password email. Therefore I changed the method to use one email address.


tomasznapieralski:
LGTM

PR #859, opened 2015-09-15
XW-305

tomasznapieralski:
https://wikia-inc.atlassian.net/browse/XW-305

PR #864, opened 2015-09-16
Fix tags for widgets tests

vforge:
/cc @tomasznapieralski @bognix @justnpT


src/test/java/com/wikia/webdriver/testcases/widgettests/GoogleFormTests.java
```diff
@@ -13,7 +13,7 @@
  */
 public class GoogleFormTests extends NewTestTemplate {
 
-    @Test
+    @Test(groups = "PollsnackWidgetTest_001")
```
bognix: how about google?

tomasznapieralski:
those changes will be introduced in
https://github.com/Wikia/selenium-tests/pull/859

PR #887, opened 2015-09-30
ADEN-2494 Check browser errors for scroll handler feature

drets:


src/test/java/com/wikia/webdriver/testcases/adstests/TestAdsBrowserError.java
```diff
@@ -0,0 +1,47 @@
...
+    adsBaseObject.getUrl(testPage);
+
+    executeScript(addErrorListenerScript());
+    executeScript(scroll());
+
+    adsBaseObject.waitForJavaScriptTruthy("window.scrollY != 0");
+    Assertion.assertEquals(executeScript(getErrors()), "");
+  }
+
+  private String addErrorListenerScript() {
```
justnpT: Have you consider moving javascript related logic to some JavaScript handler class ?
Storing scripts as String n the code doesn't make it easier to handle / modify / manage the javascripts used in the selenium repository

drets: Thanks @justnpT !

> Storing scripts as String in the code doesn't make it easier.

Completely agree with this statement.
Any suggestions how to store JS code as not a String type inside Java repo?
I can think only about using another Selenium bindings - JavaScript binding.

> Have you considered moving javascript related logic to some JavaScript handler class ?

This kind of test is unique if I may say. I would rather move JS logic to `JavascriptActions` class when we will have 2 cases where we can apply such logic.

fraszczakszymon:
LGTM

src/test/java/com/wikia/webdriver/testcases/adstests/TestAdsBrowserError.java
justnpT: I Must say the JavascriptActions class seems to be good destination for the method, as there is already isElementInViewPort method inside that class.

drets: @justnpT :+1: 
> I was thinkig about storing Javascripts in .js files and reading them in the java methods, then pushing the javascripts to execute them in the webdriven browser

I was thinking about it as well, but it will require much more work which I can't spend on this ticket. We need to read JS file with a lot of methods as a String and parse it _wisely_ and pass it to Selenium `executeScript`. If you would implement such mechanism in our repo I would be happy to use it and to move my JS logic into separate JS file.

PR #895, opened 2015-10-07
XW-112 change selector to map

bognix:
https://wikia-inc.atlassian.net/browse/XW-112

**DON'T MERGE BEFORE XW-112 IS RELEASED TO PROD ON MERCURY**


justnpT:
LGTM

PR #917, opened 2015-10-22
ADEN-2485 Compare images with different sizes

drets:
Fix the image comparison logic and make it possible to compare images with different sizes.


nandy-andy:
Hm, interesting let's put a warning and see the actual outcome.

LGTM for now :japanese_ogre: :+1:

PR #939, opened 2015-11-19
ADEN-2629 Disable all evolve tests

fraszczakszymon:


drets:
LGTM, however could you please add more context to the description section **why** we disable them?

nandy-andy:
Actually we don't need them to be disabled anymore. Closing.

PR #945, opened 2015-12-01
SOC-1728 add forgotten password test case for lowercase username

bekcunning:
This is the exact test case as SpecialPage forgotten password, but the username is lowercase in this added test.  I had to modify the string used to verify since the avatar selector has the first letter capitalized.


justnpT:
Did you consider veryfing this case on the unit test level ? I do apologize for asking such high level question but the whole functional test add's up a lot to the test suite execution time and is prone to external dependencies. It seems more sensible to verify such cases on the unit test level because this case differs from the similar one only by the username string. Thus it seems to be perfect candidate for a fast and independent unit test.

bekcunning:
No, I didn't consider that since this was a specific request for this test case.  Maybe this would be a discussion for inside the ticket at SOC-1728

ludwikkazmierczak:
LGTM

PR #950, opened 2015-12-04
ADEN-2724 Fix Ooyala selector

fraszczakszymon:
Due to last update of Chrome browser we can't easily get `object` DOM elements.


drets:
LGTM :+1:

PR #981, opened 2016-01-13
Changed seletor to match current

ludwikkazmierczak:
There is no longer an Iframe inside lightbox, so I had to change a selector to fix the lightbox test


PiotrGackowski:
LGTM

ukaka:
LGTM

PR #984, opened 2016-01-15
ADEN-2846 Update special page for videos name

nandy-andy:
`Special:Video` works but it's a redirect to `Special:Videos`. Let's remove the redirect and go straight to the right page.


drets:
LGTM :+1:

PR #985, opened 2016-01-15
ADEN-2789 Add check of Nielsen integration

drets:
Add check of Nielsen Digital Content Ratings integration.


drets:
Team decided to change tests for checking integration on wikis where Nielsen is going to be enabled (FMI ADEN-2861)

PR #986, opened 2016-01-19
ADEN-2844 Refactor extra market test

drets:
Move checking extra marker on DOM level instead of console logs.


nandy-andy:
LGTM :+1:

PR #989, opened 2016-01-21
XW-958 | update tests for new navigation on Mercury

tomasznapieralski:
https://wikia-inc.atlassian.net/browse/XW-958


src/test/java/com/wikia/webdriver/common/core/SEOUtils.java
```diff
@@ -1,7 +1,7 @@
 package com.wikia.webdriver.common.core;
 
 import com.wikia.webdriver.common.logging.PageObjectLogging;
-import com.wikia.webdriver.pageobjectsfactory.pageobject.mercury.BasePageObject;
+import com.wikia.webdriver.elements.mercury.old.BasePageObject;
```
justnpT: Why do you think it is necessary to indicate that the elememnts are "old" ?
I think that it should be removed or renamed, because reffering to time in such places can lead to misunderstanding in long term.
Eg. NewTestTemplate suffers from using new, while it is already old.

tomasznapieralski: Bring this point to QM tech meeting

tomasznapieralski:
Don't merge before:
https://github.com/Wikia/mercury/pull/1855

PR #992, opened 2016-01-22
Added a check for current email NOT equals to new email

ludwikkazmierczak:
Added a simple check if 'new email' is not the same as 'current email' - in this situation change will not trigger the mail change confirmation e-mail, and that what was causing the QAART-703 issue.


tomasznapieralski:
LGTM

justnpT:
Please check if this fixes the tests by https://wikia-inc.atlassian.net/browse/QAART-703 and if yes, remove the related issue id from the affected tests

ludwikkazmierczak:
I'm in the middle of checking that

ludwikkazmierczak:
It works on my machine, but I'm waiting for Jenkins results

PR #994, opened 2016-01-25
ADEN-2830 Add tests for Evolve2 on Mercury

drets:
Extend Evolve2 tests to Mercury skin

### DIFF
--- a/src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/adsbase/AdsEvolveObject.java
+++ b/src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/adsbase/AdsEvolveObject.java
@@ -17,6 +17,10 @@ public class AdsEvolveObject extends AdsBaseObject {
     super(driver, page, new Dimension(1366, 768));
   }
 
+  public AdsEvolveObject(WebDriver driver) {
+    super(driver);
+  }
+
   public void verifyEvolveInSlot(String slotName) {
     String slotSelector = AdsContent.getSlotSelector(slotName);
     wait.forElementPresent(By.cssSelector(slotSelector + EVOLVE_SELECTOR));
@@ -37,6 +41,12 @@ public class AdsEvolveObject extends AdsBaseObject {
     verifyNoEvolveInSlot(AdsContent.FLOATING_MEDREC);
     verifyNoEvolveInSlot(AdsContent.PREFOOTER_LEFT);
     verifyNoEvolveInSlot(AdsContent.PREFOOTER_RIGHT);
+  }
+
+  public void verifyEvolveCallMercury() {
+    verifyEvolveInSlot(AdsContent.MOBILETOP_LB);
+    verifyEvolveInSlot(AdsContent.MOBILE_AD_IN_CONTENT);
+    verifyEvolveInSlot(AdsContent.MOBILE_PREFOOTER);
   }
 
   public void verifyEvolveHoppedInSlot(String slotName, String nextProviderSrc) {
--- a/src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/adsbase/AdsEvolveObject.java
+++ b/src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/adsbase/AdsEvolveObject.java
@@ -33,7 +33,7 @@ public class AdsEvolveObject extends AdsBaseObject {
     PageObjectLogging.log("Evolve", slotSelector + " slot doesn't have Evolve.", true, driver);
   }
 
-  public void verifyEvolveCall() {
+  public void verifyEvolveCallOasis() {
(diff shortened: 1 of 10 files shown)


src/test/java/com/wikia/webdriver/testcases/adstests/TestAdsEvolveOasis.java
```diff
@@ -8,42 +8,42 @@
...
   )
-  public void testEvolveCall_CA(String wikiName, String article) {
-    String testedPage = urlBuilder.getUrlForPath(wikiName, article);
-    AdsEvolveObject wikiPage = new AdsEvolveObject(driver, testedPage);
-    wikiPage.verifyEvolveCall();
+  public void adsEvolveOasisCA(String wikiName, String article) {
+    adsEvolveOasis(wikiName, article);
   }
 
   @GeoEdgeBrowserMobProxy(country = "AU")
```
justnpT: it could be interesting to move the country codes (abbreviations) to some class, where the values could be accessed through the full country name like: countryCode.Australia. This way it would improve the understanding of test and enhance the maintenance

src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/adsbase/AdsEvolveObject.java
```diff
@@ -39,6 +43,12 @@ public void verifyEvolveCall() {
     verifyNoEvolveInSlot(AdsContent.PREFOOTER_RIGHT);
   }
 
+  public void verifyEvolveCallMercury() {
```
nandy-andy: What do you think about renaming the previous method to `verifyEvolveCallOasis()`?

src/test/java/com/wikia/webdriver/testcases/adstests/TestAdsEvolveOasis.java
nandy-andy: That's a nice idea if we use them in more tests (I assume we do). However, writing such class from a scratch is an overkill in my opinion. Maybe just finding some open-sourced solution we could re-use?

justnpT: O there is this library with CountryCode enum but this library would be an overkill in my opinion.
http://stackoverflow.com/questions/139867/is-there-an-open-source-java-enum-of-iso-3166-1-country-codes

nandy-andy: Two opinion and two overkills :smile_cat::gun: :smile_cat::gun:

drets: I analysed 2 approaches and @justnpT's approach looks "less overkill" to [me](https://github.com/Wikia/selenium-tests/blob/5313b07deb3a6a38bb438d04a2c3cfe0ea755dfd/src/test/java/com/wikia/webdriver/common/core/geoedge/CountryCode.java) :smirk_cat: 

2 points why:
* we use *only* country codes (ISO 3166-1 alpha-2 code) and country names
* country list grew only on 3-5 countries during last 1.5 years - not a big deal

Thanks @justnpT, the change truly increase readability :wink: 

---

Another *bigger* issue related to GeoEdge.
We have 2 ways to set-up GeoEdge:
* via [annotation](https://github.com/Wikia/selenium-tests/blob/5313b07deb3a6a38bb438d04a2c3cfe0ea755dfd/src/test/java/com/wikia/webdriver/testcases/adstests/TestAdsScreenshotComparisonMobile.java#L35)
* via [config](https://github.com/Wikia/selenium-tests/blob/5313b07deb3a6a38bb438d04a2c3cfe0ea755dfd/src/test/java/com/wikia/webdriver/testcases/adstests/TestAdsMonitoringOasis.java#L22)

(we need to think which approach is better and stick to one of them)

src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/adsbase/AdsEvolveObject.java
drets: You are right - it should be consistent :+1: 

*Note: I don't like the approach to keeping 2 methods related to different platforms (skins) in one class however separating them would require some changes in whole tests architecture.*

fraszczakszymon:
LGTM

justnpT:
lgtm

drets:
:whale:

src/test/java/com/wikia/webdriver/testcases/adstests/TestAdsEvolveOasis.java
nandy-andy: @drets ticket in ads automation epic, please. we can groom it and discuss the issue.

PR #995, opened 2016-01-25
Soc 1406

qaga:


src/test/java/com/wikia/webdriver/testcases/logintests/ForgottenPasswordTests.java
```diff
@@ -49,13 +52,39 @@ public void remindPasswordFromLoginDropdown() {
     dropdown.verifyUserLoggedIn(userName);
   }
 
+  @Test()
+  @Execute(onWikia = "agas")
```
justnpT: this environment should be standardized because if this wiki is not one dedicated for automated tests only (eg. manual testing is performed on that wiki as well), a different dedicated silent wiki should be used.

justnpT:
What is the stability of this test over 10 executions ?

src/test/java/com/wikia/webdriver/testcases/logintests/ForgottenPasswordTests.java
qaga: this test is not dependent on wiki data - it tests core functionality that always should work the same way regardless of wiki content.

qaga:
I run it 10 times and 10 pases

justnpT:
lgtm

PR #1000, opened 2016-01-28
XW-1045 | log user ID if possible

bognix:


justnpT:
lgtm

tomasznapieralski:
follow up in #1022

PR #1002, opened 2016-01-28
ADEN-2712 Update FMR

drets:
Since the logic of appending Floating medrec is changed we need to trigger it before looking for an element in the DOM.

**Please, don't merge it before app release (https://github.com/Wikia/app/pull/9433)**


fraszczakszymon:
LGTM :+1:

PR #1008, opened 2016-02-01
ADEN-2888 Upgrade evolve tests

nandy-andy:
Evolve2 provides more parameters in `dfp-slot-params`. We want our test to verify if the params are there as well. We had almost all pieces ready but they needed small customization:
* new data providers which include DFP client id,
* second version of `verifyGptIframe()` which as the first parameter takes the DFP client id,
* new tests run only in NZ.

### DIFF
--- a/src/test/java/com/wikia/webdriver/common/dataprovider/ads/AdsDataProvider.java
+++ b/src/test/java/com/wikia/webdriver/common/dataprovider/ads/AdsDataProvider.java
@@ -296,7 +296,6 @@ public class AdsDataProvider {
         {
             "adtest",
             "SyntheticTests/DfpParams",
-            5441,
             "wka.life/_adtest//article",
             "TOP_LEADERBOARD",
             Arrays.asList(
@@ -553,6 +552,35 @@ public class AdsDataProvider {
                 "\"loc\":\"top\"",
                 "\"pos\":\"TOP_RIGHT_BOXAD\"",
                 "\"src\":\"gpt\""
+            )
+        }
+    };
+  }
+
+  @DataProvider
+  public static Object[][] dfpEvolveParamsOasis() {
+    return new Object[][]{
+        {
+            "adtest",
+            "SyntheticTests/DfpParams",
+            4403,
+            "ev/wikia_intl/ros/TOP_LEADERBOARD",
+            "TOP_LEADERBOARD",
+            Arrays.asList(
+                "\"s0\":\"life\"",
+                "\"s1\":\"_adtest\"",
+                "\"s2\":\"article\"",
+                "\"dmn\":\"wikiacom\"",
+                "\"hostpre\":\"",
+                "\"wpage\":\"synthetictests/dfpparams\"",
+                "\"ref\":\"direct\"",
+                "\"lang\":\"en\"",
+                "\"esrb\":\"teen\""
+            ),
+            Arrays.asList(
+                "\"sect\":\"ros\"",
+                "\"pos\":\"a\"",
+                "\"site\":\"wikia_intl\""
             )
         }
     };
--- a/src/test/java/com/wikia/webdriver/common/dataprovider/ads/AdsDataProvider.java
+++ b/src/test/java/com/wikia/webdriver/common/dataprovider/ads/AdsDataProvider.java
@@ -296,6 +296,7 @@ public class AdsDataProvider {
         {
             "adtest",
             "SyntheticTests/DfpParams",
+            5441,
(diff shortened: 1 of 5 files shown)


src/test/java/com/wikia/webdriver/common/dataprovider/ads/AdsDataProvider.java
```diff
@@ -558,6 +558,35 @@ private AdsDataProvider() {
...
   @DataProvider
+  public static Object[][] dfpEvolveParamsOasis() {
+    return new Object[][]{
+        {
+            "adtest",
+            "SyntheticTests/DfpParams",
+            4403,
+            "ev/wikia_intl/ros",
+            "TOP_LEADERBOARD",
+            Arrays.asList(
```
drets: Do you think we need to test those unrelated directly to Evolve2 GPT params?
If no, change it to `Collections.emptyList()` please.

src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/adsbase/AdsBaseObject.java
```diff
@@ -36,6 +36,7 @@
   // Constants
   private static final int MIN_MIDDLE_COLOR_PAGE_WIDTH = 1600;
   private static final int PROVIDER_CHAIN_TIMEOUT_SEC = 30;
+  private static final int WIKIA_DFP_CLIENT_ID = 5441;
```
drets: :+1:

src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/adsbase/AdsBaseObject.java
```diff
@@ -332,9 +333,30 @@ public AdsBaseObject verifyProvidersChain(String slotName, String providers) {
...
    */
   public void verifyGptIframe(String adUnit, String slotName, String src) {
-    String iframeId = "google_ads_iframe_/5441/" + adUnit + "/" + src + "/" + slotName + "_0";
+    verifyGptIframe(WIKIA_DFP_CLIENT_ID, adUnit, slotName, src);
+  }
+
+  /**
+   * Test whether the correct GPT ad unit is called
+   *
+   * @param dfpClientId in most cases it's Wikia id but we have other partners like Evolve or Turtle
```
drets: :haircut: My auto formatter in IntelliJ moves *Turtle* to the next line:
```
   * @param dfpClientId in most cases it's Wikia id but we have other partners like Evolve or
   *                    Turtle
```

src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/adsbase/AdsBaseObject.java
```diff
@@ -332,9 +333,30 @@ public AdsBaseObject verifyProvidersChain(String slotName, String providers) {
...
+   * @param src the source of an ad, for example gpt, remnant or empty
+   */
+  public void verifyGptIframe(int dfpClientId, String adUnit, String slotName, String src) {
+    String iframeId;
+
+    if(src.isEmpty()) {
+      iframeId = "google_ads_iframe_/" + dfpClientId + "/" + adUnit + "/" + slotName + "_0";
+    } else {
+      iframeId = "google_ads_iframe_/" + dfpClientId + "/" + adUnit + "/" + src + "/" + slotName + "_0";
+    }
```
drets: What about get rid of this `if` and make `verifyGptIframe` even more beautiful? :smirk_cat: 

1. Change `src` param to be optional.   
It will allow to call method without `""`: `ads.verifyGptIframe(dfpClientId, adUnit, slot);`
2. Get rid of `if` and use Guava `Joiner ` instead.

```java
  public void verifyGptIframe(int dfpClientId, String adUnit, String slotName, String... src) {
    String iframeId = Joiner.on("/").skipNulls().join(
            "google_ads_iframe_",
            String.valueOf(dfpClientId),
            adUnit,
            src.length > 0 ? src[0] : null,
            slotName + "_0"
        );
    ...
}
```

src/test/java/com/wikia/webdriver/common/dataprovider/ads/AdsDataProvider.java
nandy-andy: I think it's worth checking them with and without Evolve. I'm going to leave it here ;)

drets:
LGTM :+1:

PR #1011, opened 2016-02-03
improving test case name in global navigation on mercury

justnpT:
I analyzed the test cases and improved their names as part of https://wikia-inc.atlassian.net/browse/XW-980

### DIFF
--- a/src/test/java/com/wikia/webdriver/testcases/mercurytests/NavigationTest.java
+++ b/src/test/java/com/wikia/webdriver/testcases/mercurytests/NavigationTest.java
@@ -30,7 +30,7 @@ public class NavigationTest extends NewTestTemplate {
     new Navigate(driver).toPage(MercurySubpages.MAIN_PAGE);
   }
 
-  @Test(groups = "openAndCloseNavigationAndItsSubmenu")
+  @Test(groups = "navigation-openAndCloseNavigationAndItsSubmenu")
   public void openAndCloseNavigationAndItsSubmenu() {
     init();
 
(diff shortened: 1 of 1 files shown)


vforge:
:+1:

src/test/java/com/wikia/webdriver/testcases/mercurytests/NavigationTest.java
```diff
@@ -31,7 +31,7 @@ private void init() {
   }
 
   @Test(groups = "mercury-navigation-001")
```
bognix: consider renaming groups to something more descriptive so when displayed on dashboard or in Jenkins there is more context available to the reader

justnpT: done :+1:

justnpT:
please don't merge until lgtm from @tomasznapieralski 
We have to update the Jenkins group names

tomasznapieralski:
We shouldn't remove `navigation-` prefix, because while running all tests we won't know what was tested.
Please restore prefix, we can drop numeration but we can't drop this prefix.

justnpT:
@tomasznapieralski Done

tomasznapieralski:
Thanks @justnpT good idea to add this prefix to only group name.
I will update groups on Jenkins

tomasznapieralski:
damit `mercury-` prefix shoud be also applied, i will update that

tomasznapieralski:
done, prefix added, tests updated on jenkins

PR #1019, opened 2016-02-11
Image serving rename file fix

ukaka:
Just one line fix, but I created PR for you to check if I understand what test should check correctly.

Previously:
Test was uploading image A, then renaming it to B, then trying to open not existing image A and rename it to A again.

Now:
Test is uploading image A, then renaming it to B, then opens image B and renames it to A again.


tomasznapieralski:
LGTM

PR #1040, opened 2016-02-24
Social team testcases naming

justnpT:
Changing method names in order to represent business value in user oriented way.
All the test cases are written on confluence
see: https://wikia-inc.atlassian.net/wiki/collector/pages.action?key=SOC
Tree browser -> Test cases

I did not update group names in order to:
1. Not to mess with how Jenkins executes the tests before the release
2. Because we don't yet have a clear and properly presented rule on how we name the tests


ludwikkazmierczak:
LGTM very good Job Michał

PR #1065, opened 2016-03-03
DAT-3867 add group InfoboxBuilderTests

dianafa:
We want to make this tests run in jenkins jobs.

@tomasznapieralski


src/test/java/com/wikia/webdriver/testcases/infoboxbuilder/InfoboxBuilderTests.java
```diff
@@ -14,7 +14,7 @@
 @Execute(onWikia = "mediawiki119")
 public class InfoboxBuilderTests extends NewTestTemplate {
```
tomasznapieralski: add here `@Test(groups = "InfoboxBuilderTests)"`, so you don't need to add this group for all tests

tomasznapieralski:
LGTM

PR #1066, opened 2016-03-04
Added group for every notifications tests

ukaka:
* Added common group for all notifications tests
* Added common group for all chat tests

Checked on Jenkins, pre- and post-release jobs doesn't use groups I added, so running tests won't be duplicated.


bulaszenko:
LGTM

PR #1070, opened 2016-03-08
Fix mercury tests

tomasznapieralski:
http://jenkins.wikia-prod:8080/view/Mercury%20Backgroud%20Tasks/job/Mercury-TEST-emulator/11285/

@azarek


Warkot:
I don't like it. It's good for a short term high priority test failure fix, but not as an actual solution to the problem. This shows that a user has to close the box, to see the whole footer. Maybe adding some additional page space under the footer when RWA is opened would be a way to go?

tomasznapieralski:
Yes and no. In test scenario we don't need to always do every user actions. If we want to test redirection in this test case, all we expect is link and redirection. Other test scenario should be ability to test RWA banner. We don't need to do this in one test.

Warkot:
We even shouldn't. :wink: I also heard from @azarek that the feature will be revamped. So OK, I feel convinced, LGTM. :+1:

PR #1078, opened 2016-03-11
Remove metacafe provider check

ukaka:
According to https://wikia-inc.atlassian.net/browse/SUS-170 we're retiring support for videos from metacafe.com provider, so I removed it from tests.


drets:
LGTM :+1:

PR #1091, opened 2016-03-21
XW | add more verbose comment in case of known error

bognix:


src/test/java/com/wikia/webdriver/testcases/mercurytests/old/curatedcontenttests/EditorTests.java
```diff
@@ -172,7 +172,7 @@ public void MercuryCuratedEditorTest_002_addAndSaveSection() {
     Assertion.assertNumber(
         curatedContent.getCuratedContentItemsNumber(),
         1,
-        "New section items"
+        "If error says that 3 elements were found - it means getList API returned cached response - ticket created: XW-1281"
```
Warkot: I would add there "probably", you never know. ;) And BTW, this is seen in the test's log, right?

bognix: Yes it is seen only if assertion failed.
I don't like the word probably - if I add probably then the person who checks the result is going to spend the time to investigating this issue. My intention here was to save this person time

Warkot: It's a good intention, but the road to... Never mind. It's okay for now, as it's going to be fixed in the near future, hopefully. :wink:

Warkot:
LGTM

PR #1100, opened 2016-03-31
Bumped up selenium version and Removed unused headers from Helios 

ludwikkazmierczak:
+ Bumped up selenium version to 2.53.0
+ Deleted unused headers (client_id and client_secret) from Helios call
+ HTMLUnit driver was moved to be a selenium subproject, adapted this change by adding a separate dependency

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -380,15 +380,15 @@ test {
 }
 
 dependencies {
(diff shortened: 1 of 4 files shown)


justnpT:
LGTM

PR #1110, opened 2016-04-04
ADEN-3148 Moved floor adhesion tests to project43 wiki

nandy-andy:
In order to make our synthetic tests more stable we want to run them on project43 wiki.


PiotrGackowski:
LGTM

PR #1176, opened 2016-05-04
(CE-3667) Require login for all GlobalShortcuts tests

Grunny:
GlobalShortcuts no longer loads for anon users, as such we need to ensure
the tests are run as a logged in user.

/cc @Wikia/spitfires


adamkarminski:
LGTM

PR #1182, opened 2016-05-09
ADEN-3317 Remove corporate pages test

nandy-andy:
The tests were checking if there is no Liftium on corporate pages. But we move more and more pages to `fandom.com` where Liftium is not available, therefore we decided to remove those tests.


fraszczakszymon:
LGTM 👍

PR #1199, opened 2016-05-17
ADEN-3336 Moved amazon tests on Oasis to project43

nandy-andy:
Changes:
* moved the synthetic test page to `project43` wiki,
* added a verification of bid response.


src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/adsbase/AdsAmazonObject.java
```diff
@@ -17,6 +18,7 @@
 public class AdsAmazonObject extends AdsBaseObject {
 
   private static final String AMAZON_SCRIPT_URL = "amazon-adsystem.com/e/dtb";
+  private static final String AMAZON_BID_URL = "aax.amazon-adsystem.com/e/dtb/bid";
```
drets: Could we re-use `AMAZON_SCRIPT_URL` or change `AMAZON_SCRIPT_URL` to `aax.amazon-adsystem.com/e/dtb/bid`?

drets:
Minor comment, otherwise LGTM 👍

src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/adsbase/AdsAmazonObject.java
nandy-andy: This things are two different things. That's why I decided to keep them separated. The script URL gets different JavaScript assets and it may be served from different domain in future. The bid request may also go to other domain in future and it's not for serving assets but giving us a response from online bidding auctions. Re-using the domain part now makes sense but it will add additional concatenation code. I'm not sure if we want to do it. What do you think?

drets: If it is different things - LGTM 👍

PR #1202, opened 2016-05-17
SER-80 | Disable MercurySignupTest_002_signupErrorEmailInUse while validatior is disabled

kvas-damian:
https://wikia-inc.atlassian.net/browse/SER-80

//cc @rogatty


rogatty:
LGTM, thanks

PR #1206, opened 2016-05-19
Wait for comscore to load

ukaka:
`iframe[title='VisualDNA Analytics']` element does not exists on any page. To wait for page to load I added waiting for Comscore to load as it appears on every page.


ludwikkazmierczak:
LGTM

PR #1210, opened 2016-05-19
Moved AuthModal to Lombok getter

ludwikkazmierczak:
- Moved AuthModal to Lombok getter
- code cleanup for AuthModal


ukaka:
LGTM 👍

PR #1224, opened 2016-05-24
adding related issue for a test failure

justnpT:


Archbold:
@justnpT could you use more simpler English as I can guess but i don't really know what you mean by, " _the failure is not fully repetitive,_"

Also please add whether test should be tested manually or not.

justnpT:
@Archbold  thank you for the CR fixed are in place

src/test/java/com/wikia/webdriver/testcases/forumtests/ForumBoardTests.java
```diff
@@ -85,6 +86,8 @@ public void anonymousUserCanStartDiscussionWithVideoOnForum() {
   }
 
   @Test(groups = {"ForumBoardTests_006", "ForumBoardTests", "Forum"})
+  @RelatedIssue(issueID = "MAIN-7213", comment = "the failure is caused by the environment instability. " +
```
tomasznapieralski: instead of concatenation you can move second argument to new line

justnpT:
@Archbold  Thanks that's correct I've improved the readibility. Back to you 👍

PR #1233, opened 2016-05-27
XW-1515 | review mercury tests

tomasznapieralski:
https://wikia-inc.atlassian.net/browse/XW-1515

### DIFF
--- a/src/test/java/com/wikia/webdriver/testcases/mercurytests/old/PortableInfoboxTests.java
+++ b/src/test/java/com/wikia/webdriver/testcases/mercurytests/old/PortableInfoboxTests.java
@@ -37,7 +37,6 @@ public class PortableInfoboxTests extends NewTestTemplate {
 
     // Before infobox expanding
     infobox
-        .isHeroImageNotVisible()
         .isTitleNotVisible()
         .isImageInTitleNotVisible()
         .isImageInTabberVisible()
--- a/src/test/java/com/wikia/webdriver/testcases/mercurytests/old/PortableInfoboxTests.java
+++ b/src/test/java/com/wikia/webdriver/testcases/mercurytests/old/PortableInfoboxTests.java
@@ -4,7 +4,6 @@ import com.wikia.webdriver.common.contentpatterns.MercurySubpages;
 import com.wikia.webdriver.common.contentpatterns.MercuryWikis;
 import com.wikia.webdriver.common.core.annotations.Execute;
 import com.wikia.webdriver.common.core.annotations.InBrowser;
-import com.wikia.webdriver.common.core.annotations.RelatedIssue;
 import com.wikia.webdriver.common.core.drivers.Browser;
 import com.wikia.webdriver.common.core.helpers.Emulator;
 import com.wikia.webdriver.common.templates.NewTestTemplate;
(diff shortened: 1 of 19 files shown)


src/test/java/com/wikia/webdriver/testcases/mercurytests/old/PortableInfoboxTests.java
```diff
@@ -96,18 +94,16 @@ public void mercury_infobox_externalLinkRedirectsToCorrespondingUrl() {
...
     init();
     navigate.toPage(MercurySubpages.INFOBOX_1);
 
     // Check image
     infobox
-        .clickHeroImage()
+        .clickOnImageInInfobox()
         .isLightboxOpened()
-        .closeLightbox()
-        .isTitleOverImageVisible();
```
bognix: why we don't check if title over image is visible anymore?

src/test/java/com/wikia/webdriver/testcases/mercurytests/old/widgettests/TwitterTests.java
```diff
@@ -89,12 +89,12 @@ public void MercuryTwitterWidgetTest_004_areLoadedOnFirstVisitDirectlyFromUrl()
...
   }
 
-  @Test(groups = "MercuryTwitterWidgetTest_005", enabled = false)
+  @Test(groups = "MercuryTwitterWidgetTest_005")
   public void MercuryTwitterWidgetTest_005_isErrorPresent() {
     init();
 
     widget.createIncorrect(TWITTER_INCORRECT_WIDGET_ARTICLE_NAME);
-    navigate.toPage(TWITTER_INCORRECT_WIDGET_ARTICLE_NAME);
+    navigate.toPage("/wiki/" + TWITTER_INCORRECT_WIDGET_ARTICLE_NAME);
```
bognix: why you need to append /wiki/ to this URL and you don't have to do this in the test case above?

tomasznapieralski: case above is wrapper for create article method and because all articles starts with `/wiki/` it requires to provide only article name.

the second method is universal, you have to pass whole path

src/test/java/com/wikia/webdriver/testcases/mercurytests/old/PortableInfoboxTests.java
tomasznapieralski: it looks like it was used as an indicator to check if lightbox is opened.
And because test should check that images and videos will be opened in lightbox this method is redundant

bognix: ok, but now we don't really check if lightbox is closed

tomasznapieralski: that is true, but we don't have to do it here, we have different test for checking lightbox functionality, now we want to focus on infobox only :)

tomasznapieralski: and if lighbox is not closed in that case, it will not be able to open video again, so the test will fail

bognix: I would feel better if there would be a way to verify if lightbox is close so the test would fail with proper message in that case, but it looks fine for now.

bognix:
LGTM, great work on making our tests more reliable!

tomasznapieralski:
All tests are passing on Jenkins, now we have to keep that state :)

PR #1244, opened 2016-06-03
ADEN-3427 Changed pattern for a wiki without recovery enabled

nandy-andy:
The detection domain changes from time to time and there is no reason to check exactly the domain. We just want to be sure the request was sent at all with the right param. The modification below do that.


jacekjursza:
LGTM 👍

PR #1256, opened 2016-06-09
Soc 2605

justnpT:


justnpT:
stability check: http://jenkins:8080/job/test-stability-check/68/

PR #1311, opened 2016-07-04
DAT-4564: Mock ads for global shortcuts tests

lukaszkonieczny:
Ad covers Global Shortcuts entry point in bottom Wikia bar so it cannot be clicked.

cc @Wikia/west-wing @drets @ludwikkazmierczak

### DIFF
--- a/src/test/java/com/wikia/webdriver/testcases/globalshortcuts/KeyboardShortcutsTest.java
+++ b/src/test/java/com/wikia/webdriver/testcases/globalshortcuts/KeyboardShortcutsTest.java
@@ -9,7 +9,7 @@ import com.wikia.webdriver.common.core.helpers.User;
 import com.wikia.webdriver.common.templates.NewTestTemplate;
 import com.wikia.webdriver.pageobjectsfactory.pageobject.HomePage;
 
-@Execute(asUser = User.USER, onWikia = "globalshortcuts-en")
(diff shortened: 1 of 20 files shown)


src/test/java/com/wikia/webdriver/testcases/globalshortcuts/KeyboardShortcutsTest.java
```diff
@@ -9,7 +9,7 @@
 import com.wikia.webdriver.common.templates.NewTestTemplate;
 import com.wikia.webdriver.pageobjectsfactory.pageobject.HomePage;
 
-@Execute(asUser = User.USER, onWikia = "globalshortcuts-en")
```
SebastianMarzjan: As long as ad behavior (that a closeable ad is covering a part of user functionality) is verified differently, LGTM

nandy-andy: Good point, @SebastianMarzjan! :) We've got [tests for interstitial](https://github.com/Wikia/selenium-tests/blob/master/src/test/java/com/wikia/webdriver/testcases/adstests/TestAdsInterstitial.java#L42) the most important cases for us are tested it. However, we won't mind if somebody adds to it test checking if the close button works :japanese_ogre:

PR #1312, opened 2016-07-04
DAT-4565: Mock ads for anon naviagates to RWA from menu test

lukaszkonieczny:
Ad covers whole screen so it's not possible to click menu entry point

cc @Wikia/west-wing @drets @PiotrGackowski @ludwikkazmierczak

### DIFF
--- a/src/test/java/com/wikia/webdriver/testcases/mercurytests/RecentWikiActivityTest.java
+++ b/src/test/java/com/wikia/webdriver/testcases/mercurytests/RecentWikiActivityTest.java
@@ -18,6 +18,7 @@ import org.testng.annotations.Test;
 public class RecentWikiActivityTest extends NewTestTemplate {
 
   @Test(groups = "mercury_recentWikiActivity_anonNavigateToRWAfromMenu")
+  @Execute(mockAds = "true")
   public void mercury_recentWikiActivity_anonNavigateToRWAfromMenu() {
     new MainPage()
         .open()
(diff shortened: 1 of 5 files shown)


src/test/java/com/wikia/webdriver/testcases/mercurytests/RecentWikiActivityTest.java
```diff
@@ -18,6 +18,7 @@
 public class RecentWikiActivityTest extends NewTestTemplate {
 
   @Test(groups = "mercury_recentWikiActivity_anonNavigateToRWAfromMenu")
+  @Execute(mockAds = "true")
```
SebastianMarzjan: I know that this is the interstitial, so as long as that ad behavior is verified differently, LGTM

nandy-andy: Good point, @SebastianMarzjan! :) We've got [tests for interstitial](https://github.com/Wikia/selenium-tests/blob/master/src/test/java/com/wikia/webdriver/testcases/adstests/TestAdsInterstitial.java#L42) the most important cases for us are tested it. However, we won't mind if somebody adds to it test checking if the close button works :japanese_ogre:

lukaszkonieczny: @Wikia/adeng @SebastianMarzjan tests for interstitial are now expanded on verifying does close button work.

SebastianMarzjan:
LGTM. 
@nandy-andy @lukaszkonieczny - While I usually prefer to have a single assertion per test, it looks like in this specific example any of the assertions failing would be a sign of a big problem, and splitting the assertions would increase test execution time significantly.

PR #1313, opened 2016-07-04
Change click to scrollAndClick() method

nikodamn:
The problem was caused by test waiting for click first of the pi-item, but template had many pi-title's elements on the first view. It was probably caused by some previous bug (like add title instead of data item).

I fixed the template and added changes below to prevent such situations.

Reviewers:
@Wikia/west-wing


SebastianMarzjan:
LGTM

PR #1314, opened 2016-07-04
Assert on lowercased strings

nikodamn:
Fix test by assertion on lowercased values of variables. Assertion failed when comparing links ".../fandom" with ".../Fandom".

Reviewers:
@Wikia/west-wing


idradm:
LGTM

PR #1319, opened 2016-07-05
Dat 4591 - split groups into smaller ones

nikodamn:
Split groups:
- PortableInfoboxTests
- InfoboxBuilderTests

into smaller groups. It will increase stability of these tests. Now, unfortunately, whole group is send to one Jenkins' VM which for larger group of tests (like these mentioned above) is causing a lot of failures related to errors of execution.

Reviewers:
@Wikia/west-wing


rybmat:
LG

PR #1340, opened 2016-07-15
SUS-662 escape values of HTML attributes

mixth-sense:
The test failed while comparing a string with its representation as a value of an HTML attribute.

The expected value was `"Download" Adult Coloring Book Stress Relieving Cats PDF Free`.

The actual value was `&quot;Download&quot; Adult Coloring Book Stress Relieving Cats PDF Free`.

It is a defect of the test not the code being tested. The expected value should be first escaped.

@ludwikkazmierczak


src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/BasePageObject.java
```diff
@@ -318,7 +319,7 @@ public void waitForValueToBePresentInElementsAttributeByCss(String selector, Str
     changeImplicitWait(250, TimeUnit.MILLISECONDS);
     try {
       waitFor.until(CommonExpectedConditions.valueToBePresentInElementsAttribute(
-          By.cssSelector(selector), attribute, value));
+          By.cssSelector(selector), attribute, escapeHtml(value)));
```
ludwikkazmierczak: I don't think we should escape on this level, we should explicitly escape title of a video when passing to a method, not in here. Let's move it to `verifyVideoAdded(String videoTitle)` in `SpecialVideosPageObject`

mixth-sense:
@ludwikkazmierczak — please do not comment on outdated diffs.

ludwikkazmierczak:
LGTM

PR #1367, opened 2016-07-22
Add checking 5th and 6th page views

nikodamn:
To test which is checking if Sales Pitch Dialog is visible on 4th page view add verifying if it is NOT visible on 5th and 6th page view.

Reviewers:
@Wikia/west-wing


SebastianMarzjan:
LGTM ( though I'm wondering if using a sequence of articles instead of the same one would cover the problem more fully )

PR #1395, opened 2016-07-29
XW-1581 | fixes due to Design System footer changes

Warkot:
@Wikia/x-wing

I'll look if there are more changes needed. Anyway, we have to merge it do master, not earlier and not later than when realising the footer itself.


tomasznapieralski:
LGTM

PR #1396, opened 2016-07-29
ADEN-3644 Fix for monitoring and infobox change

nandy-andy:
After the change which moved `MOBILE_TOP_LEADERBOARD` below Portable Infobox (PI) our monitoring tests started to fail on some wikis (those with PI enabled).


jacekjursza:
LGTM

PR #1429, opened 2016-08-30
ADEN-3495 check IV3 in different way

jacekjursza:


src/test/java/com/wikia/webdriver/testcases/adstests/TestIVWAnalyticsProvider.java
```diff
@@ -51,11 +52,10 @@ public void testIV3AnalyticsProviderInDE(String wikiName, String path) {
       networkTrafficInterceptor.startIntercepting();
       String testedPage = urlBuilder.getUrlForPath(wikiName, path);
       AdsBaseObject adsBaseObject = new AdsBaseObject(driver, testedPage);
+      JavascriptActions jsActions = new JavascriptActions(driver);
 
-      assertTrackingPixels(adsBaseObject,
-                           URL_BASE_SCRIPT,
-                           URL_TRACKING_SCRIPT);
-
+      jsActions.waitForJavaScriptTruthy("typeof(iom) == 'object'");
```
bognix: I never knew you can do `typeof(...)` although it's a little misleading because `typeof` is not a function, it's an operand, adding `()` makes it look like it's a function.

How about triple `=`?

bognix:
Minor comment, besides LGTM.

Did you create a ticket in QAART project about the problems with multiple searching inside `HAR` files?

PR #1431, opened 2016-08-30
Mercury suggestions fix

nikodamn:
Test was failing because it was trying to test Mercury global nav on Oasis skin (Discussion page) on wiki without Discussions feature. I added annotation to use "dauto" wiki, also created a sample article there.

Reviewers:
@ludwikkazmierczak


ludwikkazmierczak:
LGTM

PR #1437, opened 2016-09-02
Removed method checking not existing feature

ukaka:
https://wikia-inc.atlassian.net/browse/SUS-756

Removed method that checks sorting dropdown on Special:Videos page as this feature was removed months ago.


nikodamn:
LGTM

PR #1440, opened 2016-09-06
ADEN-3530 fix flite test

jacekjursza:
Changed CSS selector in data provider (the div IDs are kind of random).


PiotrGackowski:
LGTM

PR #1447, opened 2016-09-07
SUS-789 Remove support for GameStar video provider

ukaka:
https://wikia-inc.atlassian.net/browse/SUS-789

There is a ticket to sunset GameStar video provider support (https://wikia-inc.atlassian.net/browse/SUS-979) and we're not going to fix it, so I removed the test which check GameStar video adding.


nikodamn:
LGTM

PR #1454, opened 2016-09-09
ADEN Set mockAds cookie on correct domain

fraszczakszymon:
Set mockAds cookie on main domain instead of subdomain


jacekjursza:
LGTM

jacekjursza:
LGTM (after talking to @ludwikkazmierczak )

PR #1456, opened 2016-09-11
ADEN-3941 Adjust data provider to WikiFactory changes

nandy-andy:
Categories for project43.wikia.com has been changed. Below changes fix it in our tests.


PiotrGackowski:
LGMT

PR #1463, opened 2016-09-13
ADEN-3941 Move Ooyala pre-roll test to project43

nandy-andy:
Moved the test from adtest.wikia.com to project43.wikia.com and renamed it to follow [our convention](https://wikia-inc.atlassian.net/wiki/display/ADEN/Ad+Engineering+Automated+Test+Naming+Convention).


bulaszenko:
LGTM

PR #1484, opened 2016-09-21
WW-95 go to login page if anon editing disabled tests added

idradm:
Please do not merge before https://github.com/Wikia/app/pull/11352 is on production

### DIFF
--- a/src/test/java/com/wikia/webdriver/testcases/communitypagetests/SpecialCommunityPageTests.java
+++ b/src/test/java/com/wikia/webdriver/testcases/communitypagetests/SpecialCommunityPageTests.java
@@ -52,7 +52,7 @@ public class SpecialCommunityPageTests extends NewTestTemplate {
 
   private String getEditUrl(String page, String query) {
     return urlBuilder.appendQueryStringToURL(
-        urlBuilder.getUrlForPage(String.format("/wiki/%s", page.replace(" ", "_"))),
+        urlBuilder.getUrlForPage(String.format("/wiki/%s", urlBuilder.normalizePageName(page))),
         query);
   }
 
--- a/src/test/java/com/wikia/webdriver/testcases/communitypagetests/SpecialCommunityPageTests.java
+++ b/src/test/java/com/wikia/webdriver/testcases/communitypagetests/SpecialCommunityPageTests.java
@@ -17,7 +17,7 @@ public class SpecialCommunityPageTests extends NewTestTemplate {
   public void verifyAnonEditingForbiddenFlow() {
     SpecialCommunity page = new SpecialCommunity();
     // click the first link from any card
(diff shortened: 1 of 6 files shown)


src/test/java/com/wikia/webdriver/testcases/communitypagetests/SpecialCommunityPageTests.java
```diff
@@ -0,0 +1,58 @@
...
+public class SpecialCommunityPageTests extends NewTestTemplate {
+
+  @Execute(onWikia = "muppet", asUser = User.ANONYMOUS, mockAds = "true")
+  public void verifyAnonEditingForbiddenFlow() {
+    SpecialCommunity page = new SpecialCommunity();
+    // click the first link from any card
+    WebElement link = page.getLinkFromCards();
+    String url = getEditUrl(link.getText(), "veaction=edit");
+    link.click();
+    // should be at register page
```
sebastianmarzjan: How do we check that? It's not clear

src/test/java/com/wikia/webdriver/testcases/communitypagetests/SpecialCommunityPageTests.java
```diff
@@ -0,0 +1,58 @@
...
+    String url = getEditUrl(link.getText(), "action=edit");
+
+    link.click();
+
+    Assertion.assertEquals(url, page.getCurrentUrl());
+  }
+
+  private String getEditUrl(String page, String query) {
+    return urlBuilder.appendQueryStringToURL(
+        urlBuilder.getUrlForPage(String.format("/wiki/%s", page.replace(" ", "_"))),
```
sebastianmarzjan: What do you think about moving the normalization logic to 
getUrlForPage(String pageName) ?

src/test/java/com/wikia/webdriver/testcases/communitypagetests/SpecialCommunityPageTests.java
idradm: We do not. The test should fail if it's not true.

lukaszkonieczny: +1 Can we verify we are on register page. Also do we want to cover case when new user is joining community?

sebastianmarzjan: Are you sure? loginAs from WikiBasePageObject uses Helios to log in, what is the connection with register / login page ?

idradm: I will add assert for it to make sure.

PR #1485, opened 2016-09-22
Change REDIRECT_LINK to a page that always exists

macbre:
Do not assume that "Formatting" article always exists. Main_Page is present on all wikis and can be safely used when testing link suggest.

http://jenkins.wikia-prod:8080/view/Staging-test/view/staging/job/staging-forum-staging/lastSuccessfulBuild/artifact/logs/log.html

@ludwikkazmierczak / @nikodamn


nikodamn:
LGTM :+1:

PR #1487, opened 2016-09-23
ADEN-4006 Make Taboola tests more stable

nandy-andy:
* test only containers we control,
* check if loader is requested only when ads are blocked,
* code clean-up.


jacekjursza:
LGTM

PR #1500, opened 2016-10-03
The VisualEditorEditing_002_delete was failing because of spacing issue

sebastianmarzjan:
(removal of leading space that was not contained in the string to remove).
This fixes the issue by correcting the selection range and including
the space in string to remove.


ludwikkazmierczak:
LGTM

PR #1513, opened 2016-10-12
Revert "SOC-3137 adding RelatedIssues"

nikodamn:
Reverts Wikia/selenium-tests#1502 per @PiotrGackowski request


PiotrGackowski:
SOC-3137 is fixed

PR #1518, opened 2016-10-14
Fix call logging

sebastianmarzjan:
Removing declarations of static variables that are already defined in parent class. This fixes recording of URL in the ApiCall.call() method


ludwikkazmierczak:
LGTM

PR #1528, opened 2016-10-19
SUS-1185 Fix for VideosPageTest_002 test

bulaszenko:
Fix VideosPageTest_002 selenium test which is failing because of assertion, test is compering not removed video title with not added video tile

### DIFF
--- a/src/test/java/com/wikia/webdriver/testcases/specialpagestests/VideosPageTests.java
+++ b/src/test/java/com/wikia/webdriver/testcases/specialpagestests/VideosPageTests.java
@@ -15,6 +15,7 @@ import org.testng.annotations.Test;
 public class VideosPageTests extends NewTestTemplate {
 
   static final String VIDEO_QUERY = "truth";
+  static String ADDED_VIDEO_TITLE;
 
   /**
    * Verify UI elements on the Special:Videos page Logged-Out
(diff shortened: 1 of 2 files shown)


src/test/java/com/wikia/webdriver/testcases/specialpagestests/VideosPageTests.java
```diff
@@ -44,12 +45,17 @@ public void VideosPageTest_002() {
     YoutubeVideo video = YoutubeVideoProvider.getLatestVideoForQuery(VIDEO_QUERY);
 
     specialVideos.addVideoViaAjax(video.getUrl());
+
+    specialVideos.isNewVideoAdded();
+
+    ADDED_VIDEO_TITLE = specialVideos.getNewestVideoTitle();
```
macbre: Local method variable should be enough (instead of a static class member)

src/test/java/com/wikia/webdriver/testcases/specialpagestests/VideosPageTests.java
```diff
@@ -68,11 +74,16 @@ public void VideosPageTest_003() {
...
```
macbre: Local method variable should be enough (instead of a static class member)

macbre:
LGTM

PR #1529, opened 2016-10-19
WW-430 | Uncollapsing wikia bar before clicking links on it

rybmat:
https://wikia-inc.atlassian.net/browse/WW-430

//cc @Wikia/west-wing


lukaszkonieczny:
LGTM

PR #1530, opened 2016-10-20
Cleaning out codee for CateegoryPageObject

ludwikkazmierczak:


src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/category/CategoryPageObject.java
```diff
@@ -1,51 +1,42 @@
...
+  /**
+   * Get image URL for a specific article from gallery
+   * 
+   * @param articleName - name of the article
+   * @return null if there is no such article name displayed
+   */
+  public String getImageURLFromGallery(String articleName) {
+    for (WebElement page : categoryGalleryItems) {
+      if (page.findElement(By.cssSelector("a")).getAttribute("title").equals(articleName)) {
+        return page.findElement(By.cssSelector("img")).getAttribute("src");
```
idradm: What happens if there is an element without an image?

src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/category/CategoryPageObject.java
```diff
@@ -1,51 +1,42 @@
...
-            }
-        }
-        return index;
-    }
-
-    public String getPageImageURL(int index) {
-        wait.forElementVisible(categoryGalleryItemImages.get(index));
-        return categoryGalleryItemImages.get(index).getAttribute("src");
-    }
+    return null;
```
idradm: Have considered using exception here instead?

src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/category/CategoryPageObject.java
ludwikkazmierczak: Just "elementNofFoundException" and test is stoped

src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/category/CategoryPageObject.java
ludwikkazmierczak: i don't consider it as a good pattern, IMHO it will only complicate, cause we will have to handle exception in the test - which would overcomplicate test flow

PR #1542, opened 2016-10-26
SOC-3306

PiotrGackowski:
rebranding change


PiotrGackowski:
Wait with merge until name of fallout app will be renamed on iOS store.

PR #1544, opened 2016-10-27
SOC-3314

PiotrGackowski:
Removing RelatedIssues for test that check functionality will be fixed on 11/3


PiotrGackowski:
Don't merge till SOC-3151 will go to production.

PR #1555, opened 2016-11-03
ADEN-4038 stabilize recovery test

jacekjursza:
Sometimes we get error:
`org.openqa.selenium.WebDriverException: unknown error: window._sp_.getElementId is not a function`
it might be because of SourcePoint lib is not loaded on time, this change should fix it


nandy-andy:
LGTM :+1:

PR #1559, opened 2016-11-04
Assert that there are five hero units

JoshuaRogan:
Check to see if there are five clickable units in the hero mosaic.

### DIFF
--- a/src/test/java/com/wikia/webdriver/elements/fandom/components/HeroBlock.java
+++ b/src/test/java/com/wikia/webdriver/elements/fandom/components/HeroBlock.java
@@ -1,5 +1,6 @@
 package com.wikia.webdriver.elements.fandom.components;
 
+import org.openqa.selenium.By;
 import org.openqa.selenium.WebElement;
 import org.openqa.selenium.support.FindBy;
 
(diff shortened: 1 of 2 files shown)


src/test/java/com/wikia/webdriver/elements/fandom/components/HeroBlock.java
```diff
@@ -19,4 +20,8 @@ public boolean isDisplayed() {
       return false;
     }
   }
+
+  public boolean hasFiveHeroUnits() {
+    return heroMosaic.findElements(By.cssSelector(".clickable-unit")).size() == 5;
```
JoshuaRogan: Is there a cleaner way to do this with the decorator?

mklucsarits: You could try:
```
@FindAll({@FindBy(css = "section.hero-block .hero")})
private List<WebElements> heroUnits;
```

But I think it's fine doing it this way. Only thing I would change is selecting on `.hero`, because `clickable-unit` is used for tracking and doesn't necessarily define the element itself. That CSS class could change or move to another element inside the hero unit.

mklucsarits:
LGTM

PR #1593, opened 2016-11-25
SUS-1151-b add DontRun annotation to two Mercury_Search tests

nikodamn:
* Add `DontRun` annotation and remove these tests from Mercury_Search_002 group (due to found bug with `DontRun` annotation blocking whole test's group).

Reviewers:
@mixth-sense


mixth-sense:
Tested and accepted. :medal_sports:

PR #1601, opened 2016-11-30
added Related Issue to failing infobox test

qaga:


qaga:
No need to add RI as functionality was fixed meanwhile

PR #1602, opened 2016-12-01
Fix next button on the second step of Create New Wiki process

lukaszkonieczny:
We style buttons to be design system compliant so this button has additional classes, not only `next`

@Wikia/west-wing @mixth-sense @ludwikkazmierczak @nikodamn


lukaszkonieczny:
CNW test passes http://jenkins:8080/view/parametersTesting/job/run-test-group/264/

PR #1632, opened 2016-12-30
SOC | turn off test because of timeout

mleczey:
## Description
Test for following on following tab timeouts because of huge amount of posts. Pagination will be introduced in SOC-3674.
## Reviewers
@tortila


src/test/java/com/wikia/webdriver/testcases/discussions/FollowingPostTests.java
```diff
@@ -211,9 +213,10 @@ public void userOnDesktopCanUnfollowPostOnUserPostsPage() {
   /**
    * By default all posts on "Followed" tab are followed.
    */
-  @Test(groups = "discussions-userDesktopFollowingPost")
+  @Test(groups = "discussions-userDesktopFollowingPost", enabled = false)
```
tortila: can you please add "TODO" in a comment with information when to enable these tests again?

mleczey: Added comment to SOC-3674, discussed that this will be enough.

PR #1636, opened 2017-01-05
ADEN-4427 replace HOME_TOP_LEADERBOARD with TOP_LEADERBOARD

dianafa:
Adeng introduces change that removes `HOME_` prefix from adslots

https://wikia-inc.atlassian.net/browse/ADEN-4427


src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/wikipage/WikiArticleHomePage.java
```diff
@@ -6,7 +6,7 @@
 
 public class WikiArticleHomePage extends WikiBasePageObject {
 
-  private By wikiHomePageSpecificElement = By.cssSelector("#HOME_TOP_LEADERBOARD");
+  private By wikiHomePageSpecificElement = By.cssSelector(".mainpage");
```
nandy-andy: I want to suggest running some other tests which use it since I know AdEng tests passed but let's make sure other selenium tests passes. The worry comes from the fact that the selector used to be a unique id `#HOME_TOP_LEADERBOARD` and now it's going to be changed to a class `.mainpage`. The adeng tests may not be used `wikiHomePageSpecificElement` and passed whereas the `By.cssSelector()` may now return a collection and cause failures in some other tests. What do you think about this concern?

dianafa: Thanks for sharing this! `wikiHomePageSpecificElement` is used in`wait.forElementPresent()` which in the end calls `findElement(locator, driver)`.  `findElement` method returns a WebElement object otherwise it throws an exception - so it always returns a first found element - simply checks for presence so any cssSelector is safe here. I've however changed `cssSelector` to `className` to be more fancy ;) all tests passes on jenkins.

PR #1646, opened 2017-01-17
SOC-3861 fix invalid permissions for tests

tortila:
Ticket: https://wikia-inc.atlassian.net/browse/SOC-3861

This should fix most of the selenium tests failures on jenkins


slayful:
LGTM, great catch!

PR #1716, opened 2017-03-13
Avatar - findElement removed

sczerwinski-wikia:


src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/WikiBasePageObject.java
```diff
@@ -388,8 +388,8 @@ public void verifyUserLoggedIn(final String userName) {
         wait.forElementClickable(MERCURY_NAV_ICON);
         driver.findElement(MERCURY_NAV_ICON).click();
       } else {
+        wait.forElementVisible(By.cssSelector(LOGGED_IN_USER_SELECTOR_OASIS));
```
tortila: `wait.forElementVisible` returns `WebElement`, so you can assign the value to `avatar` here and then remove the line below completely

PR #1706, opened 2017-03-02
SUS-1737: In Wall and Comments tests double-check that user is logged in

TK-999:
https://wikia-inc.atlassian.net/browse/SUS-1737

PR #1760, opened 2017-04-11
Revert "SUS-1737: In Wall and Comments tests double-check that user is logged in"

mixth-sense:
Reverts Wikia/selenium-tests#1706


mixth-sense:
Per conversation with @tortila and @TK-999, the changed test introduced performance problems when a sequence of tests is executed.

PR #1922, opened 2017-07-27
Deleted everything connected to Maps feature

ludwikkazmierczak:
Deleted everything connected to Maps feature

@macbre


macbre:
The less Java the better :)

PR #1955, opened 2017-09-05
GRADLE updated to 4.1

TK-999:


gradle.properties
```diff
@@ -0,0 +1,3 @@
+org.gradle.daemon=true
+org.gradle.configureondemand=true
```
ludwikkazmierczak: Deamon and configuration on demand caused problems before, i think we should keep them false

PR #2058, opened 2017-10-31
ADEN-6133 Removed Ooyala tests

nandy-andy:


src/test/java/com/wikia/webdriver/testcases/adstests/TestAdsOoyalaOasis.java
```diff
@@ -1,22 +0,0 @@
-package com.wikia.webdriver.testcases.adstests;
-
-import com.wikia.webdriver.common.dataprovider.ads.AdsDataProvider;
-import com.wikia.webdriver.common.templates.TemplateNoFirstLoad;
-import com.wikia.webdriver.pageobjectsfactory.pageobject.adsbase.AdsOoyalaObject;
```
nandy-andy: We need to keep it because of mobile-wiki tests

PR #2063, opened 2017-11-02
ADEN-5961 Add A9 video tests

v3rth:


src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/adsbase/AdsAmazonObject.java
```diff
@@ -1,9 +1,30 @@
 package com.wikia.webdriver.pageobjectsfactory.pageobject.adsbase;
 
+import com.wikia.webdriver.common.contentpatterns.AdsContent;
 import org.junit.Assert;
 import org.openqa.selenium.WebDriver;
 
 public class AdsAmazonObject extends AdsBaseObject {
+  private static final String QS_TURN_ON_A9 = "InstantGlobals.wgAdDriverA9BidderCountries=[XX]";
```
fraszczakszymon: I have mixed fillings about putting those consts in this class. Shouldn't page object represent specific page/DOM state?

src/test/java/com/wikia/webdriver/testcases/adstests/TestAmazonAds.java
```diff
@@ -1,57 +1,56 @@
...
     for (String slotName : slots) {
-      amazonAds.verifyTestBidAdInSlot(slotName, A9_TEST_LINE_ITEM);
+      amazonAds.verifyTestBidAdInSlot(slotName, AdsAmazonObject.A9_TEST_LINE_ITEM);
     }
   }
+
+  private void testAmazonVideo() {
+    networkTrafficInterceptor.startIntercepting();
+    AdsAmazonObject amazonAds = new AdsAmazonObject(driver, AdsDataProvider.PAGE_FV.getUrl(AdsAmazonObject.A9_VIDEO_DEBUG_MODE));
+    amazonAds.wait.forSuccessfulResponseByUrlPattern(networkTrafficInterceptor, AdsAmazonObject.A9_VIDEO_DEBUG_BID_PATTERN);
```
fraszczakszymon: I'm not sure if I understand it correctly - it checks whether we requested bid only, right? Wouldn't it be better to check whether we passed bid parameter to VAST request *and* check whether ad is playing? You can reuse `waitForAdPlaying`

https://github.com/Wikia/selenium-tests/blob/master/src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/adsbase/AdsJWPlayerObject.java#L37

src/test/java/com/wikia/webdriver/pageobjectsfactory/pageobject/adsbase/AdsAmazonObject.java
v3rth: I agree, it should... but it isn't :) For us `Page` object is more like "ad product" from test perspective and it keeps everything what you need to test specific (in this case for example run debug mode for A9, verify if there is bid in slot etc.) functions. It should be thinner as you said, but in that case we need typical ad-product object with functions specific for every ad-product (run FV video, unmute VUAP video) but I think that would be too complicated and this way is simpler.

PR #2104, opened 2017-12-06
SUS-3503 Prepare UserAndRights tests

mszabo-wikia:
Cleanup `UserAndRights` tests so they can be run.

https://wikia-inc.atlassian.net/browse/SUS-3503


nikodamn:
Is it still valid or we can close it?

PR #2174, opened 2018-02-12
Video adaptation

chessky:


src/test/java/com/wikia/webdriver/common/dataprovider/VideoUrlProvider.java
```diff
@@ -10,12 +10,7 @@ private VideoUrlProvider() {
   @DataProvider
   private static final Object[][] videoUrl() {
     return new Object[][]{
-        // Premium Provider Links
-        {
-            //ign
```
tortila: If we keep ign support, then we should keep a test for this provider - in this case please just find any video and replace the link in this data provider. Otherwise we can remove it

chessky: The provider is up to be sunset and it is not possible to add video from them - so we can remove it.

PR #2241, opened 2018-04-05
IRIS-5781 Guidelines default text for newly created Discussions page

nikodamn:
IRIS-5781 Add test for guidelines default text for newly created Discussions page.

@Wikia/iris


src/test/java/com/wikia/webdriver/testcases/discussions/DiscussionsCreation.java
```diff
@@ -23,6 +24,15 @@ public void userCanSeeEmptyDiscussionsMessageAndAddPostOnNewCommunityDiscussions
     Assertion.assertFalse(page.getPost().isPostListEmpty());
   }
 
+  @Execute(asUser = User.USER_CNW)
+  @InBrowser(emulator = Emulator.DESKTOP_BREAKPOINT_BIG)
+  public void guidelinesHaveDefaultContentOnNewCommunityDiscussionsPage() {
```
rogatty: The name is a bit misleading. We only test if it's not empty.

PR #2266, opened 2018-05-02
ADEN-7030 Add verification to featured, inline and video page slots

dmnsobczak:


src/test/java/com/wikia/webdriver/testcases/adstests/TestAdsFandomArticleVideo.java
```diff
@@ -68,4 +72,16 @@ public void adsArticleVideoAdsMobile() {
   public void adsArticleVideoMOATTrackingMobile() {
     adsArticleVideoMOATTrackingDesktop();
   }
+
+  private void verifySlots(AdsFandomObject fandomPage, Boolean skipTopLeaderboard) {
+    fandomPage.triggerOnScrollSlots();
+
+    if (!skipTopLeaderboard) {
```
nandy-andy: Is it because there is no TLB on a video page?

src/test/java/com/wikia/webdriver/testcases/adstests/TestAdsFandomVideoPage.java
```diff
@@ -0,0 +1,48 @@
...
+    verifySlots(pageObject);
+  }
+
+  @InBrowser(
+      browser = Browser.CHROME,
+      emulator = Emulator.GOOGLE_NEXUS_5
+  )
+  @Test(groups = "AdsVideoPageF2Mobile")
+  public void adsVideoPageAdsMobile() {
+    adsVideoPageAdsDesktop();
```
nandy-andy: This doesn't look right: a test for mobile devices uses method called `adsVideoPageAdsDesktop` ;) How about making `adsVideoPageAdsDesktop` general and renaming it to `adsVideoPageAds` and then having two proxy methods with the correct annotations for desktop and mobile?

src/test/java/com/wikia/webdriver/testcases/adstests/TestAdsFandomArticleVideo.java
dmnsobczak: Good suggestion, changed.

src/test/java/com/wikia/webdriver/testcases/adstests/TestAdsFandomVideoPage.java
dmnsobczak: I based on standards in other tests but yeah, it's always great moment to start changing somethings :) Renamed.

PR #2440, opened 2018-09-10
ADEN-7653 Change of line item ID

emadry:


src/test/java/com/wikia/webdriver/common/dataprovider/ads/AdsDataProvider.java
```diff
@@ -403,7 +403,7 @@ private AdsDataProvider() {}
                                                        AdsContent.MEDREC
                                                    )
                                                        .put("slotSize", new Dimension(300, 250))
-                                                       .put("lineItemId", "365404452")
+                                                       .put("lineItemId", "365416332")
```
nandy-andy: That's because non-SRA was enabled site-wide! 😈

PR #2513, opened 2018-11-05
Small fixes for tests

swietlana:
- small change in pattern (just space) for weird chars in post tests
- added scrolling to elements (guidelines, explore tests)
- changed german wiki for trending articles test on mobile

PR #2516, opened 2018-11-13
IW test fixes

swietlana:


src/test/java/com/wikia/webdriver/elements/communities/desktop/components/navigation/local/CommunityHeaderDesktop.java
```diff
@@ -146,7 +146,7 @@ public CommunityHeaderDesktop openExploreMenu() {
 
   public SpecialWikiActivityPageObject clickExploreWikiActivityLink() {
     wait.forElementClickable(exploreWikiActivityLink);
-    scrollAndClick(exploreWikiActivityLink);
```
rybmat: why are we replacing scrollAndClick with just click?

swietlana: Sometimes page was scrolled before clicking the exploreTab and that was the reason test failed. So I've changed before for (https://github.com/Wikia/selenium-tests/pull/2513/files) `scrollTo(exploreTab)` but it worked only locally. Now it should work as scrolling to exploreTab and just choosing the option e.g. Videos

